PR #2, opened 2011-06-19
Adding README.md for Github description

natecavanaugh:
When viewing a project on Github, it will parse a README file to allow a description of the project to be used. This normally includes instructions on how to get up and running with the project, links to other resources, etc.

I'm adding this in so we have something rather than the note that we don't have a README.

We should add more info to this, but just wanted to get it in.


jackelyneaquino:
jackelyneaquino

PR #485, opened 2013-11-23
LPS-42559: Enabled generic URI path parameter stripping

lucastheisen:
The current code just removed jsessionid, however, that is just on example of path parameter.  Further the regex replace assumed that anything after jsessionid was throw away.  While that may be the case in the current Tomcat implementation, path parameters are allowed on any path segment, and not just jsessionid.  In order to be more compliant with the servlet spec, I updated the FilterMapping to strip all uri parameters.  

I would actually suggest doing this before getting into FilterMapping, like in InvokerFilterHelper before the loop, so that it could be done once per request rather than once per filter per request, but not sure if you guys had any other reason for doing it this way.  If you comment back, i can move the code so it happens in InvokerFilter before merging.

https://issues.liferay.com/browse/LPS-42559


inacionery:

inacionery:

PR #577, opened 2014-06-02
LPS-44990 Provide ability to deploy Faceted search implementations as a hook

tdetoux:
Implement the solution suggested by Ray Auge :
- A new facet configuration parameter is read from the facet's JSON string : pluginContext.
- If plugin context is not empty, we look for the facet implementation in the plugin's classLoader.


inacionery:

inacionery:

sammonsjl:
ci:close

PR #655, opened 2014-12-18
LPS-52283 : Asset Publisher scope group lost while publishing from remot...

tdetoux:
Here is a proposed solution for this problem :
- During lar export, scope group ids are converted to friendlyURLs
- During lar import, frienlyURLs are converted back to scope group ids

Hence, the asset publisher will keep its original configuration as long as exists a site withe the same friendlyURL.


inacionery:

sammonsjl:

sammonsjl:
ci:close

PR #659, opened 2015-01-02
LPS-52186 Added a basic confirmation dialog to satisfy feature request

javeedchida:
My changes in sitemap.jsp started out as a modification of the existing aui:a tag whereby I set its href to a javascript invocation. My tests revealed Firefox doesn't play well with that. So I went with the more proper approach of pining the href attribute to "#" and adding an onclick event to the aui:a tag. But then I ran into a problem with aui:a not respecting my "return false" to prevent the default behavior of the link. So I used a straight html anchor tag and augmented it to do what aui:a was doing for me, namely: added a span to address the "opens in a new window" icon and a title attribute for accessibility.

If there is a way to still use aui:a, I'll be glad to improve this solution.


javeedchida:
Newbie stuff...this pull request was created off of master. Just created a branch and re-created a pull request off of that branch.

PR #660, opened 2015-01-02
LPS-52186 Defined a basic confirmation dialog for sitemap submission

javeedchida:
My changes in sitemap.jsp started out as a modification of the existing aui:a tag whereby I set its href to a javascript invocation. A quick revealed Firefox doesn't play well with that, so I went with the more proper approach of pinning down the href attribute to "#" and adding an onclick event to the aui:a tag. But then I ran into a problem with aui:a not respecting my javascript's "return false" to prevent the default behavior of the link. So I used a straight html anchor tag and augmented it to do what aui:a was doing for me, namely: added a span to address the "opens in a new window" icon and a title attribute for accessibility.

If there is a way to still use aui:a, please share, and I'll be glad to improve this solution.


inacionery:

inacionery:

sammonsjl:

sammonsjl:
ci:close

PR #687, opened 2015-03-13
LPS-54266 available_languages.jsp has wrong content type on Wildfly 8.2

deleze:
Using response header manipulation doesn't work on Wildfly 8.2 with Java 8: Content-type is text/hrml.

```
response.addHeader(HttpHeaders.CONTENT_TYPE, ContentTypes.TEXT_JAVASCRIPT);
```

The header is overridden and then Chrome doesn't accept to execute the javascript.
Use this will fixe the content type and wildly respond with "text/javascript".

It is better to set the JSP page setting instead of manipulating directly the response header.


inacionery:

inacionery:

sammonsjl:

sammonsjl:
ci:close

PR #695, opened 2015-03-28
When an Action method raises an exception, forward that exception rather...

lolo101:
... than an irrelevant NoSuchMethodException.

When developing a portal contained in Liferay, we had some hard time debugging our portal because the logged exception was a NSME whereas the action method was actually called and the actual exception thrown by the action method was trashed.


lolo101:
Invalidated by commit
https://github.com/liferay/liferay-portal/commit/37001513b2e5c7da45839c1948983053c40612ca

PR #701, opened 2015-04-10
LPS-54883: ServiceBuilder currently broken; hardcoded "/src/"

FuadEfendi:
LPS-54883: ServiceBuilder cannot retrieve JavaClass from "/service/" locations; generated implementations do not contain necessary methods.


FuadEfendi:
I made trivial error in this initial "pull" request.

Suggested code change fixing the issue, see line 4406 of ServiceBuilder.java:

BEFORE:

		int pos = fileName.lastIndexOf("/src/") + 5;

AFTER:

		int pos = fileName.lastIndexOf("/src/");
		if (pos == -1) {
		    pos = fileName.lastIndexOf("/WEB-INF/service/") + 17;
		} else {
		    pos = pos + 5;
		}

Plus, if some crazy developer uses package name such as "my.package.src.MyClass" (with "src") we have troubles again :)

FuadEfendi:
Not applicable anymore; closing this.

PR #706, opened 2015-04-20
Lps 55084 Move HibernatePortalCacheManger out of LiferayEhcacheRegionFactory, prepare for extracting portal cache

tinatian:
@shuyangzhou 
Hi Michael,

This is the first group of changes for cleaning before extracting portal cache as a bundle. 

I made these changes because I want to move HibernatePortalCacheManager into portal-cache-ehcache bundle instead of portal-hibernate-ehcache, and make portal-hibernate-ehcache a client of HibernatePortalCacheManager service.  
The benefits are 1) do not need to export implementation details(for example EhcachePortalCacheManager) 2) make HibernatePortalCacheManager consistent with other portal cache manager services

In case you want to see the whole picture, please see https://github.com/mhan810/liferay-portal/pull/349

Thanks.
Tina.


tinatian:
close

PR #724, opened 2015-05-13
Lps 55562 Make cache reconfigurable by osgi bundle

tinatian:
@shuyangzhou 

Hi Michael,

I add support to make cache reconfigurable by osgi bundle, and also keep it reconfigurable by deploying plugins, we can remove the old way once everything is moved to bundle, please check.

Thanks.
Tina.


tinatian:
close

PR #734, opened 2015-06-01
Lps 55790 Create new LockService interface and replace all references to LockLocalService and LockLocalServiceUtil with new service interface

tinatian:
@shuyangzhou 

Hi Michael,

This is the first ticket about lock modularization, please check.

I also have finished the second ticket (move lock implementation out), will commit after this one is merged.

I found some bugs about cache, cluster and scheduler while testing the lock, will work on them while you are reviewing the code.

Thanks.
Tina.


tinatian:
close

PR #755, opened 2015-07-27
[TECHNICAL-SUPPORT] LPS-56374 Set staging life cycle

moltam89:
Hey Máté,

even though in **LayoutRemoteStagingBackgroundTaskExecutor** the **ExportImportThreadLocal.setLayoutStagingInProcess(true);** is called, but it affects only the export side, and in case of remote staging, the import side is unaware of this flag.

Actually, this is a separate bug from [LPS-56374](https://issues.liferay.com/browse/LPS-56374), however the fix for LPS-56374 depends on it, so I think the best solution would be to commit this fix under LPS-56374 as well, so we will have a complete solution.

Thanks,
Tamás


moltam89:
Sent here by accident.

PR #764, opened 2015-08-11
LPS 57820

marcoscv-work:
Fixing product menu blink effect on load through open and closed classes


marcoscv-work:
ouch, wrong pull !

PR #779, opened 2015-09-07
LPS-58412

planetsizebrain:
Added priority to updateAsset in Journal Service
Changed priority from Integer to Double in Asset Service
Added priority.jsp and associated properties

### DIFF
--- a/modules/apps/asset/asset-publisher-test/test/integration/com/liferay/asset/publisher/service/test/AssetPublisherServiceTest.java
+++ b/modules/apps/asset/asset-publisher-test/test/integration/com/liferay/asset/publisher/service/test/AssetPublisherServiceTest.java
@@ -211,7 +211,7 @@ public class AssetPublisherServiceTest {
 
 			JournalArticleLocalServiceUtil.updateAsset(
 				TestPropsValues.getUserId(), article, assetCategoryIds,
-				assetTagNames, null);
+				assetTagNames, null, null);
 
 			AssetEntry assetEntry = AssetEntryLocalServiceUtil.getEntry(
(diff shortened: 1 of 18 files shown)


planetsizebrain:
@juliocamarero here you can find my first attempt for integrating my priority hook stuff directly into Liferay. I guess my formatting and imports style might still be wrong in places. The same updateAsset change should also work for other places it seems like bookmarks etc...

juliocamarero:
Hey Jan (@planetsizebrain),

this is really awesome. I will try to merge this today. I just need you to accept our "Contribution Agreement" here: https://issues.liferay.com/browse/LPS-58412  (Login, and then Workflow --> Contribute Solution)

Thanks a lot!!

juliocamarero:
Hey Jan (@planetsizebrain),

I just sent your pull for the final review, you can see it here: https://github.com/brianchandotcom/liferay-portal/pull/30012

I split your commit in smaller ones and then added some minor changes after each commit so that you could easily review what I have changed. I have also tried to explain the reason for every change in the commit message but feel free to ask me anything or let me know if you don't agree with any of those changes and we can talk about them.

Your pull was really good! Thanks a ton for contributing to the project! :+1:

planetsizebrain:
Hi @juliocamarero,

Your comment made me learn something again (that the serviceContext could be even exploited further) and I think adding the priority to categorisation is indeed better (I had added it separately because I wasn't sure). 

Happy to contribute!

juliocamarero:
thanks a lot for you words @planetsizebrain, I added it to ServiceContext because now we can easily update any other asset to use this without having to modify the API :)

Hopefully this should be merged by tomorrow!

PR #806, opened 2015-10-30
LPS-59249 - Form-groups shouldn't be nested within each other, remove uneeded form styles

pat270:
Hey Chema, Sending this to you because I removed form-control from liferay_ui["input-editor"] unless its alloy-editor. It was causing the DDM HTML Input's container to be fixed height. See https://github.com/pat270/liferay-portal/commit/81f53d6e7c205ea2c78c9bbfb3946f3d6d51f829.

Also the style of alloy editor will change due to the form-control class, alloy editor in styled theme will have borders. Nate said this might go against the style you guys were trying to achieve. See screenshot.
screen shot 2015-10-30 at 2 49 49 pm


jbalsas:
Hey @pat270, I guess you sent this to liferay by mistake, could you please forward it to me?

You guys make a good point, sorry for messing other editors up :disappointed:  I'm fine with the editor having borders in the `_styled` theme. No harm done :wink: 

I think rather than hardcoding it in the `.ftl` we could move that logic to the `InputEditorTag`. Having a explicit reference to a specific editor implementation doesn't always sound good. Send this to me and I'll take care of that.

jbalsas:
On the other hand, I think I added some custom css to make AlloyEditor's `height: auto`. Maybe it would make sense to just add it in general for any other input editor rather than removing the css class. 

I think it makes sense that something inside a form would be marked as a `form-control`.

What do you think?

PR #810, opened 2015-11-12
[Technical Support] LPS-59252 - Inconsistent locale between URL and page content

zsigmondrab:
Hi Julio,

I am sending this to you as you reviewed the original fix for the problem also. See https://github.com/juliocamarero/liferay-portal/pull/3708

This pull request is about extending that fix for handling the url with no group and layout friendly urls in the request uri and the test case.

I have tested all of the aspects, even with different context and proxy path so I can say that this modification should not cause any problem.

I have found only one problem which was a hidden one that occurred only after our fix. That is the third commit about which could be handled with a different LPS also. Please, let me know if we should do so.

Let me know that also if you have any concern.

Thank you,
Zsigmond


zsigmondrab:
Wrong repository. Sorry.

Zsigmond

PR #812, opened 2015-11-16
LPS-59313 - Blogs title extends beyond portlet boundary

daviddotzhang:


blzaugg:
@daviddotzhang, this should have been sent to me.

PR #838, opened 2016-01-04
LPS-61133 Facebook Connect VerifyProcess

stian-sigvartsen:
Hi Mike. This verify process is for the fixed Struts impl of Facebook Connect. Once we migrate away from Struts, we need to make another VerifyProcess which alters the value of oauthRedirectURL.


stian-sigvartsen:
Wrong repo!

PR #842, opened 2016-01-06
[Technica Support] LPS-60792 Reverting falling back to the previous scope type

zsigmondrab:
Hi Máté,

I could not found better solution for this because as I can see, there is no other way to handle different data level scope from the portlet level scope within a portlet.

Because of that, I only reverted the fallback logic.

Please, let me know if you have any better idea.

See the antecedent pull request: https://github.com/matethurzo/liferay-portal/pull/1273#issuecomment-162601686

Thank you,
Zsigmond


zsigmondrab:
wrong branch

PR #845, opened 2016-01-12
LPS-61962 Fix Facebook's struts action (broke again)

stian-sigvartsen:
Hi Tomas. The Facebook integration needs a slight tweak since LPS-61464 (the original Struts fix) because the Struts action needs to be associate with a portlet. I've now set it to "/portal" which is what Google SSO is also using.

There must have been a core bug that previously allowed the Facebook Struts action to work without a portlet being associated!


stian-sigvartsen:
Wrong fork. Closing.

PR #851, opened 2016-01-19
LPS-62144 Redirect after a FastPortlet successful login ActionRequest

stian-sigvartsen:
Hi Tomas. As discussed, we need to issue a redirect after a successful login request which has no redirect parameter. I originally implemented this into SecurityPortletContainerWrapper like we discussed, but then discovered there was already code in LoginMVCActionCommand specifically for handling successful logins by FastLoginPortlet. So I re-wrote to provide a more targeted fix there will less overall portal impact. Happy to change it if you prefer the original approach. Thanks!


stian-sigvartsen:
Wrong fork, closing...

PR #861, opened 2016-01-28
LPS-62312 Restoring a previous version of a document inside a folder that has restricted document type does not convert document's type

cleydyr:


cleydyr:
Sent by mistake.

PR #911, opened 2016-04-06
[TECHNICAL SUPPORT] LPS-63922 User-specific portlet preferences are not cleaned up when the portlet is removed and portlet can not be re-added 

lipusz:
https://issues.liferay.com/browse/LPS-63922

Hi Ray,

See Norbert's comment:
>Use cases to consider:
1. Preferences-unique-per-layout=true preferences-owned-by-group=true
plid+portletId will be unique in this case and that record should be removed.
2. Preferences-unique-per-layout=true preferences-owned-by-group=false
This is what the issue is about. In this case there will be one record for the layout owned by the group, and one owned by each user who configures the portlet. Every one of those needs to be removed.
3. The column is customizable.
plid+portletId will be unique (portletId + userId separator + userId) in this case and that record has to be romoved

>In other cases the preferences won't be related to the layout so they won't be removed.

I've asked him to create a test for this. We'll submit if you agree with his changes.

Thanks,
Tibor

/cc @NorbertKocsis


lipusz:
Wrong user... :-)

PR #939, opened 2016-05-10
Staging TLD description update

jhinkey:
https://issues.liferay.com/browse/LRDOCS-2559

cc/ @codyhoag


jhinkey:
there are changes in other TLDs that need to be done. I'll resend.

PR #1046, opened 2016-10-13
LPS-67407 Update the com-liferay-journal repository

juliocamarero:


brianchandotcom:
@pyoo47 @jhendley we really need a retry for the specific batch. Our pass rate is about 20% right now... adding the retry would weed out the flakes and bring it to 90-100%

brianchandotcom:
ci:merge:force

PR #1059, opened 2016-10-25
[TECHNICAL SUPPORT] LPS-65498 'Unable to encrypt blank password' error on scheduled import from LDAP

NorbertKocsis:
Hi,

Sending this again as it will be needed to be resolved in order to be able to reproduce an other issue.

To summarize again: This fix has nothing to do with changing whether or not we allow blank password for users. It just removes related logic from LDAP module (if it gets a password it tries to set it, if it doesn't get one then it doesn't) and leaves this decision entirely to the User service. (Which by the way still doesn't allow blank passwords.)

Br,
Norbert


NorbertKocsis:
Sorry! Sent to liferay/master by mistake.

Br,
Norbert

PR #1208, opened 2017-02-23
LPS-70549 & LPS-70386

mbowerman:


jpince:
@mbowerman I see at least some of these commits in master, so I'm assuming that whatever needed to go in already has.. closing this pull

PR #1224, opened 2017-03-03
LPS-42478 Set context classloader to null to make hibernate use its own classloader rather than context classloader

tinatian:
Hi Shuyang,

I tested this change locally and passed, let's see if there is any ci failure.

cc @istvansajtos @lipusz  This is another solution for this issue.

Tina


tinatian:
ci: stop

PR #1468, opened 2017-05-16
Start converting kernel.io package to a petra module

jrao:
Start converting kernel.io package to a petra module:

Removed dependencies like Liferay's Log class, StringPool, CharPool, etc.
Copied some Liferay kernel classes to the petra module for use as internal classes to break dependencies
Copied kernel.nio packages to petra.io module and made some progress breaking those dependencies


jrao:
Meant to send to @rotty3000

PR #1595, opened 2017-06-18
LPS-73153 Leave micro digit in schema version for those upgrade processes we need to backpor

achaparro:
Hi @natocesarrego,

For the same reason than here:
https://github.com/natocesarrego/liferay-portal/pull/84

We need to change the schema version for some upgrade processes in master.

Please, let me know if you have any question regarding it.

Thanks.
@josemariaMunoz


achaparro:
Wrong pull

PR #1635, opened 2017-06-28
Update JournalArticleLocalServiceImpl.java

russkayairina:
For emailType requested should be used the right email subject and body


jkappler:
@russkayairina, thanks for the change! Please see: https://github.com/liferay/com-liferay-journal/pull/378

Since it's a separate repo, your commit will be first applied to liferay-journal, then portal, and afterwards it will be backported to ee7.0.x

jkappler:
ci:close

PR #1659, opened 2017-07-04
LPS-55244 Removing ResourceBlockPermission orphane rows

mario83:


portal-impl/src/com/liferay/portal/service/impl/ResourceBlockLocalServiceImpl.java
```diff
@@ -1107,4 +1109,4 @@ protected void updatePermissionsHash(ResourceBlock resourceBlock) {
 	private static final Log _log = LogFactoryUtil.getLog(
 		ResourceBlockLocalServiceImpl.class);
 
-}
\ No newline at end of file
+}
```
sammso, line 1112: @mario83 by looking the code this is not correct fix as you are removing resource block. Y

mario83: thanks @sammso, I found the mistake.

mario83:
don't fix correctly LPS-55244

PR #1689, opened 2017-07-13
LPS-73608 - Use S3 Store to connect to another S3 compatible object storage

luneo7:
Today it is only possible to use Amazon S3 with S3Store. It would be nice to add configuration parameters to make it possible to use another object storage service that is compatible with S3 specs, for example CEPH.
To use it correctly it is necessary to make possible to set the following parameters: endpoint, protocol, path-style, signer override.

https://issues.liferay.com/browse/LPS-73608


ajcrowe:
+1

This would be very useful as it could allow for Google Cloud Storage to be configured as a backend as well.

sammonsjl:

holatuwol:
ci:reopen

holatuwol:
To make it easier to apply, strip out all the language changes.

```bash
git show --name-only ceph~1..ceph | grep -vF Language | xargs git format-patch ceph~1..ceph --
sed -i.bak 's@/apps/foundation/@/apps/@g' *.patch
```

Add a line break to `@Meta.AD(description = "bucket-name-help", name = "bucket-name")` so that the closing parentheses exists on its own line, and make that a commit. This allows us to use `git am` to capture the generated `.patch` file.

```bash
ls -1 *.patch | sort -u | xargs git am
```

Remove the line break, make that another commit. Then squash all three commits together.

holatuwol:
https://github.com/jonathanmccann/liferay-portal/pull/1818

PR #1828, opened 2017-08-23
LRQA-31884 Implement event-based flow control in build object model

pyoo47:
https://issues.liferay.com/browse/LRQA-31884

Sorry this took so long to review. I got waylaid a long time back by various issues and it took me a while to get back to it. I've made a lot of changes. We can go over my changes in person but in the meanwhile, can you test to see if I broke anything?


kevin-yen:
@pyoo47  I think you intended to send it to me instead of liferay user

PR #1884, opened 2017-09-06
LPS-47165 Prevent spurious database updates by cloned UnicodeProperties

liefke:
Extend LinkedHashMap instead of HashMap to prevent Hibernate from writing instances of these properties again and again.


tinatian:
Hi @liefke ,
This is not needed since toString now is always sorted. It was a problem when you found this, but we fixed this in LPS-47533 (d22e5ea12a17bd68fb680e5a723128d66ec55e60).

Thanks for finding this and sending us a pull.

Tina.

PR #2153, opened 2017-10-25
LPS-75477

ricardocousolr:
The attribute name should be the same as that set in com.liferay.taglib.ui.SearchPaginatorTag


ricardocousolr:
ci:retest

PR #2288, opened 2017-11-29
LPS-75107

adolfopa:


adolfopa:
Sent to the wrong user. See https://github.com/sergiogonzalez/liferay-portal/pull/4143

PR #2299, opened 2017-11-30
LRDOCS-4077 instructions for using separate read and read-write data sources

jhinkey:


jpince:
@jhinkey did you mean to send this to Brian?
also, looks like this has a merge conflict with `portal-impl/src/portal.properties`

PR #2324, opened 2017-12-05
LPS-72947

holatuwol:
Simplified one of the queries in order to see if it would pass Sybase. Discovered through additional testing that it wouldn't run on MySQL 5.6 and fixed that. Now resubmitting to CI to see if Sybase now works. If it doesn't work, I'll get a Sybase VM or container.


holatuwol:
Wrong repository.

PR #2524, opened 2018-01-11
LPS-76988 Fix doImportMissingReference methods to prevent NullPointerExceptions

moltam89:
Hey Máté,

I realized that about half of the _doImportMissingReference_ methods are missing the null checks, which can be a problem e.g. if we create REFERENCE_TYPE_WEAK references therefore it is not guaranteed that the referenced content really exists at the import side.

I'm going to send pull request to the dynamic-data-mapping and to the journal subrepo as well.

Thanks,
Tamás


jpince:
@moltam89 this was sent to the liferay user, did you mean to send it to Mate instead?

moltam89:
Hi @jpince ,

yeah, I'll resend it to Máté.

Thanks!

PR #2594, opened 2018-01-26
LPS-76828 Layout Import does not replace @ variables in URLs correctly

moltam89:


moltam89:
https://github.com/matethurzo/liferay-portal/pull/1900

PR #2603, opened 2018-01-30
LPS-77485 - Portlet 3.0 TCK: Failure in V2PortletTests that call

vsingleton:
portletConfig.getContainerRuntimeOptions


vsingleton:
opened PR to incorrect integrator.

PR #2610, opened 2018-02-01
LPS-77702 Fixed Saving Timers in Condition Node for Kaleo Workflow

brianikim:
Hello @gregory-bretall,

https://issues.liferay.com/browse/LPS-77702
https://issues.liferay.com/browse/LPP-28795

The issue is that the user is unable to save a timer on a condition node in the Liferay Kaleo workflow engine. 
The reason for this issue is that the condition element in definitions\liferay-workflow-definition_7_0_0.xsd does not have an implementation of timers in it. 
The fix for this issue is simple in that I just brought over the existing timer element from the task element into the condition element. I also made the timer element optional in the condition node so that users can decide whether they want timers saved or not.

Please let me know if there is any issue or you have any questions. Thanks!


brianikim:
Please disregard this PR. I have sent it to the wrong individual.

PR #2729, opened 2018-04-02
LPS-79313

Alec-Shay:
Relevant tickets:

https://issues.liferay.com/browse/LPP-29528
https://issues.liferay.com/browse/LPS-79313

If two bundles of Liferay are configured to use virtual hosts with different domain names, then depending on the properties used, Liferay may fail to replace URL links when exporting and re-importing a LAR with web content between them.

The main issue appears to be that Liferay is ignoring the configured values for the port, making it exit early from `LayoutReferencesExportImportContentProcessor.replaceExportHostname()` (which is supposed to replace the host name in the URL). Replacing this allows the domain name of the URL to be replaced, which fixes this issue.

### DIFF
--- a/portal-impl/src/com/liferay/portal/util/PortalImpl.java
+++ b/portal-impl/src/com/liferay/portal/util/PortalImpl.java
@@ -4072,7 +4072,12 @@ public class PortalImpl implements Portal {
 		}
 
 		if (inetSocketAddress == null) {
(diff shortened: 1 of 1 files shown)


Alec-Shay:
ci:close

Alec-Shay:
Sent to liferay user by mistake

PR #2738, opened 2018-04-09
LPS-78466 Broken hyperlinks in basic webcontent after import

moltam89:
Hey Máté,

before this fix if we had a link from a different group, we appended the group's UUID.
This works fine while staging is enabled, but causes problems when we export/import the Sites, as the UUIDs won't match between the export and the import sides.

My idea was to use the group's friendly URL instead.
Vendel made sure that the new import logic will handle the previously created LAR files as well by fetching the group by the UUID first (which is needed for staging as well).

Thanks
Tamás


moltam89:
Sent to liferay user by accident.

PR #2749, opened 2018-04-17
LPS-76333 Include the WSDD jars when running build-app-lpkg

CAustin582:


jpince:
@CAustin582 did you mean to send this to Brian?

PR #2932, opened 2018-07-02
LPS-82717

brianikim:
Hello @SamZiemer ,

Sending a new PR from here: https://github.com/SamZiemer/liferay-portal/pull/461.

> 
> Hello @SamZiemer ,
> 
> https://issues.liferay.com/browse/LPS-82717
> 
> The issue is that DLAppUtil's getExtension is passing in the Title to be checked for an extension if a sourceFile does not exist. The logic here is flawed because a Title of a document entry should never be used for checking extensions because no source Files exist. Because the logic exists, it is causing variety of issues that are described in LPS-82717.
> 
> The simple fix is to remove the use of Title being checked for extensions and/or mimeTypes.
> 
> Sincerely,
> Brian Kim


brianikim:
wrong send.

PR #2974, opened 2018-07-17
[TS] LPS-83223

yuhai:
Hi @matethurzo

The PR extends from https://github.com/matethurzo/liferay-portal/pull/2185

Please refer to the fix explanation:

Commit 1 to 2:
The root issue is ChangesetPortlet property definition wrong. Please refer to https://github.com/liferay/liferay-portal/blob/master/definitions/liferay-portlet-app_7_1_0.dtd#L478-L485. For ChangesetPortlet, its preferences-unique-per-layout should false. Otherwise we will use same the data "com_liferay_exportimport_web_portlet_ChangesetPortlet created, with ownerId = 0, ownerType = 3, plid = 0" for different groups.

Commit 3 to 13:
The root issue is we treated the portlet **com_liferay_document_library_web_portlet_DLPortlet** as DataSiteLevelPortlets. However, it shouldn't belong to this level because the portlet can be placed in layout and it is instanceable portlet.

Commit 14(the last commit).
We refer to the link from https://dev.liferay.com/en/discover/portal/-/knowledge_base/7-0/exporting-importing-app-data

> You can export/import app content two ways. You can navigate to the app’s administrative area located in the Menu, or you can visit the indiviual app that resides on a Liferay page. Both export/import menus work the same, but the administrative area may hold different content than its indiviual app counterpart

I think the statement is conflict. For now, DLAdminPortlet and DLPortlet hold same data for export/import. On 7.0, we can export/import data from DLPortlet. This is different from 6.2. For the link statement, we added the last commit to keep the current behavior, even though the method is opposite with the portlet's function. If the expected behavior is DLPortlet should not own export/import data function, please remove the commit.

cc @arthurchan35 

Regards,
Hai

### DIFF
--- a/modules/apps/blogs/blogs-test/src/testIntegration/java/com/liferay/blogs/internal/exportimport/test/BlogsExportImportTest.java
+++ b/modules/apps/blogs/blogs-test/src/testIntegration/java/com/liferay/blogs/internal/exportimport/test/BlogsExportImportTest.java
@@ -55,7 +55,7 @@ public class BlogsExportImportTest extends BasePortletExportImportTestCase {
 
 	@Override
 	public String getPortletId() {
-		return BlogsPortletKeys.BLOGS;
+		return BlogsPortletKeys.BLOGS_ADMIN;
 	}
 
 	@Before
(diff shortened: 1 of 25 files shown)


yuhai:
sorry, send wrong user. :)

PR #3007, opened 2018-08-10
LPS-84243 Re-enable Form Page controls after exiting edit mode

joshuacords:
https://issues.liferay.com/browse/LPS-84243

The page addition controls should be disabled while editing a translation per https://issues.liferay.com/browse/LPS-79730, however they should be re-enabled after leaving that mode. Also, the control trigger selection did not account that there is actually 3 available controls, 1 for a single page and 2 (initially hidden) for the multiple page header. The 2 for the multiple page header were not selected with  ".form-builder-page-header .form-builder-controls-trigger".

Note: If any changes need to be made, make sure to test both with 1 Form page and multiple Form pages.


joshuacords:
ci:test:relevant

joshuacords:
Resent here: https://github.com/pedroqueiroz/liferay-portal/pull/173

PR #3106, opened 2018-09-20
LPS-85584 Fix creation of Enum values for JSONWS arguments

igr:


adolfopa:
Hey @igr,

It is failing for me when using enums from a module. What I've done is the following:

1. Add two methods to BlogsEntryServiceImpl

```java
	@Override
	public void passEnumArg(Isolation arg) {
	}

	@Override
	public Isolation returnEnumValue() {
		return Isolation.COUNTER;
	}
```

2. Rebuild services and deploy blogs-api and blogs-service modules.
3. From the JSONWS console, send a request for the `returnEnumValue` method (this works fine)
4. From the JSONWS console, send a request to the `passEnumArg` using `COUNTER` as its value. It fails, complaining about not being able to resolve the value:

    [JSONWebServiceServiceAction:116] COUNTER cannot be found by com.liferay.blogs.service_3.0.0

adolfopa:
See https://github.com/sergiogonzalez/liferay-portal/pull/4607

PR #3202, opened 2018-10-26
LPS-86460 update db schema with PID changes 

ajsampang67:
Since this was found to be reproducible in master, continuing from https://github.com/achaparro/liferay-portal/pull/241#issuecomment-433373318

For RSS, PID was changed in configurationId but not dictionary.
For Flags, neither were changed, so added this to upgrade process

Also updated expected schema version in Flags/PageFlags portlets

cc/ @MichaelPrigge @ChrisKian


ajsampang67:
Wrong repo

PR #3341, opened 2019-01-22
LRQA-46287 ADD a LRLogin.path file

louis-zhang18:
ADD a LRTermOfUse.path file
           ADD a LRPasswordReminder.path file
           ADD a LRWelcome.path file
           Modify the MyFirstTest.testcase


louis-zhang18:
Admin error

PR #3348, opened 2019-01-29
LPS-86681 PasswordModifiedFilter check is unnecessary during imperson…

brianikim:
Hello @topolik ,

Resending from https://github.com/topolik/liferay-portal/pull/692

Impersonation check is done by comparing the request user id with the session user id as the session still maintains the impersonator's user id. We need to check whether session user id is null as well for when there is no impersonator, ex. when a user is simply just logging in.

Please let me know if you have any questions. Thanks.

Sincerely,
Brian Kim


brianikim:
Disregard. Wrong send.

PR #3360, opened 2019-02-05
LPS-89918 Cannot import page templates with pages based on them in the same LAR

moltam89:
From @Alec-Shay :

Relevant tickets:

https://issues.liferay.com/browse/LPP-32655
https://issues.liferay.com/browse/LPS-89918

If a site or site template contains a page template with one or more pages based on it, then they cannot all be imported into another site with a single import, despite all being present in the LAR.

This occurs because of a validation step that occurs for all page templates, which was necessary in previous versions when page templates could not be scoped to the site, but is now unnecessary since they can be included in the same import.

As discussed on [PTR-566](https://issues.liferay.com/browse/PTR-566), although it's not the most robust solution, we are for now going with a solution of adding a flag for layouts on import to skip template validation if they are not based on a Global page template, since it should not be essential for them, anyway.

### DIFF
--- a/modules/apps/export-import/export-import-service/src/main/java/com/liferay/exportimport/internal/controller/LayoutImportController.java
+++ b/modules/apps/export-import/export-import-service/src/main/java/com/liferay/exportimport/internal/controller/LayoutImportController.java
@@ -1372,6 +1372,14 @@ public class LayoutImportController implements ImportController {
 				continue;
 			}
 
+			boolean globalLayoutPrototype = GetterUtil.getBoolean(
+				layoutElement.attributeValue(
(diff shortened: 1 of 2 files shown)


moltam89:
ci:test:relevant

Alec-Shay:
Hello,

Sorry I didn't notice until now, but I think it was a mistake this was sent to the liferay user, so that may be why this hasn't been reviewed yet, since it's in review with Máté at the moment. I just went ahead and forwarded the same branch to him and am updating the LPS: https://github.com/matethurzo/liferay-portal/pull/2416

moltam89:
My bad, thank you @Alec-Shay !

PR #3369, opened 2019-02-13
Lps 88984c

SpencerWoo:
https://github.com/jrao/liferay-portal/pull/40

@jrao 
Code should be the same as yours except changed the initial listener function to be allPortletsReady.

portletReady is to handle the changing configuration use case -- though one could argue for refresh required

It's a shame it runs so frequently but can't seem to improve


SpencerWoo:
ci:test:sf

PR #3392, opened 2019-03-04
LPS-88743 Circularly referenced web-content throws exception on import

ricky-pan:
See conversation here: https://github.com/SpencerWoo/liferay-portal-ee/pull/68

Bug cannot be fully reproduced on master because of https://issues.liferay.com/browse/LPS-90401. However, based on the logic of the code, the issue present in 7.1 should then become reproducible after fixing LPS-90401.


ricky-pan:
ci:test:relevant

PR #3393, opened 2019-03-04
[TECHNICAL SUPPORT] LPS-90727 XSS Vulnerability in Control Panel 

knchau:
XSS vulnerability is caused be javascript appearing before the query string. I escaped the URL string so that it can be safely set as the URL attribute and the pop-up no longer appears. It escapes the string following the domain.

https://issues.liferay.com/browse/LPS-90727


knchau:
ci:test:sf

knchau:
ci

PR #3398, opened 2019-03-06
LPS-90927 Use the default language from the nameXML and compare that …

SamZiemer:
After upgrading from 7.0 fix pack 46 and lower while using Japanese as a default language, a warning will be thrown upon startup while trying to add a duplicate layout prototype. This is due to a translation change in LPS-77699.

This fix uses the default language identified by the XML so the name can be compared without translation issues.


ealonso:
Just started reviewing :) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

ealonso:
ci:close

PR #3445, opened 2019-04-20
LPS-94383 As a developer I can use Searcher to count matches only, without returning documents

arboliveira:


arboliveira:
ci:test:sf:norebase

arboliveira:
ci:test:search:norebase

arboliveira:
ci:stop

PR #3452, opened 2019-05-03
Adding Tamil language as "Community Contributed" to portal

dgomezg:
@drewbrokke  Please check if with these changes you are able to run the language builder and have Tamil finally added to the portal.


dgomezg:
//cc @JorgeFerrer

PR #3453, opened 2019-05-03
LPS-94598  Create JournalArticleMultiLanguageSearchJapaneseSummaryTest

vagnerbc:
https://issues.liferay.com/browse/LPS-94598

- Content used: **新規作成**
- Highlith used: **新規**
- com.liferay.portal.kernel.search.highlight.HighlightUtil linha: 88 
      Regex that was created in the method highlight to get the highlight term used:   **\Q新規\E\b**

* The problem is that Japanese is written / read from right to left and has no space character.
* The **"\ b"** of the regex causes the word **"新規"** to not be found

**The possible solution is to add a locale parameter in the HighlightUtil.highlight () method and check if the language code is Japan or another that has the same respects (Chinese, etc.) to not add the bondary in the regex.**


vagnerbc:
ci:test:sf

vagnerbc:
ci:test:search

PR #3574, opened 2019-09-30
_http and _layoutLocalService should be 'protected' for subclass

mcgitty:
The other 3 service @Reference have public or protected setter, which subclass can use to inject service reference to the base class JournalConverterImpl. A subclass cannot do that for these 2 private references. At the minimum, declare them protected.


sammonsjl:
ci:close

PR #3575, opened 2019-09-30
Use the 'localizable' flag to avoid exploding the XML

mcgitty:
When one field has multi-language values, all fields are duplicated by the number of "available-locales" of the root element. For example, a HTML field with images encoded in "data" URL scheme (RFC-2397) may be language neutral and very large. Duplicating it in all available languages is meaningless, wasteful and bad for performance.

The DDM structure already has the 'localizable' flag. If a field is marked as "localizable: false", this change makes JournalConverter not to duplicate the content into all available languages.


sammonsjl:
ci:close

PR #3578, opened 2019-10-03
COMMERCE-1999 fixed cache related bug on old management bar

FabioDiegoMastrorilli:
This is related to a bug we found in commerce migrating to 7.2

The management bar we've used doesn't correctly handle cached components state and this breaks at page loading.


ethanbustad:
Sent to bchan's origin here: https://github.com/brianchandotcom/liferay-portal/pull/79162.

PR #3650, opened 2019-12-12
Optimize countFieldRepetition() for getContent()

mcgitty:
getContent() will take 30+ minutes to run on a 2024 field Web Content, because each call to countFieldRepetition() repeats the same call to getDDMFieldsDisplayValues(). This optimization calls getDDMFieldsDisplayValues() only once per getContent(), and the same 2024 field Web Content takes about 7 seconds.


sammonsjl:
ci:close

PR #3715, opened 2020-02-13
LPS-107191 Deploying recharts module on ant all

andresfulla:
We were not deploying `frontend-js-recharts` when _ant-all_ing.
It's not a big deal because the only module using it, `analytics-reports-web`, is not deployed with the ant all.

This tries to solve that.

Does this make any sense.

cc/ @darquesdev @wincent


wincent:
I don't really know how this file is supposed to work, @andresfulla. Almost all the lines I looked at seemed to be autogenerated. @jbalsas can probably provide you with a useful answer.

andresfulla:
This was not the way to do it

andresfulla:
https://github.com/jbalsas/liferay-portal/pull/2144

PR #3716, opened 2020-02-14
Fix Notifications Configuration for unmodifiable Delivery Types

tbigby-kristin:
When you create a custom UserNotificationDeliveryType which has `modifiable = false` and `defaultValue = true`, these show as turned on but not user modifiable in the Configuration for the Notifications portlet.

However, on saving the configuration form, the UserNotificationDeliveryType gets turned off, because the Liferay form has POSTed the delivery type ID but no checkbox value `true` as the form input is disabled. This should not happen as the user was not meant to be able to modify the delivery type.

It's difficult to explain, but the backend code is at: https://github.com/liferay/liferay-portal/blob/7.2.x/modules/apps/notifications/notifications-web/src/main/java/com/liferay/notifications/web/internal/portlet/NotificationsPortlet.java#L301

It loops through all delivery type IDs that were displayed on the page, gets the value for each delivery type ID, and updates that value into the DB. When a delivery type ID is not modifiable, its input checkbox is disabled, and therefore will never POST a `true` value even if it is turned on. Therefore the result is that it gets saved to a value of `false` no matter what the user wanted nor what they were allowed to do.

This patch checks whether the UserNotificationDeliveryType is modifiable before adding in to the list of delivery type IDs. Therefore only the delivery types that are modifiable will be added to the list, so only the modifiable ones can be updated in the DB.

Note that this fix helps against the normal user case; however an advanced user manipulating the POST response will still be able to change a non-modifiable UserNotificationDeliveryType. A backend fix will also be necessary to prevent that - where the UserNotificationDeliveryType `isModifiable()` method is checked prior to calling `_updateUserNotificationDelivery()`.

### DIFF
--- a/modules/apps/notifications/notifications-web/src/main/resources/META-INF/resources/notifications/configuration.jsp
+++ b/modules/apps/notifications/notifications-web/src/main/resources/META-INF/resources/notifications/configuration.jsp
@@ -111,8 +111,8 @@
 												UserNotificationDelivery userNotificationDelivery = UserNotificationDeliveryLocalServiceUtil.getUserNotificationDelivery(themeDisplay.getUserId(), entry.getKey(), userNotificationDefinition.getClassNameId(), userNotificationDefinition.getNotificationType(), userNotificationDeliveryType.getType(), userNotificationDeliveryType.isDefault());
 
 												if(userNotificationDeliveryType.isModifiable()) {
-                          userNotificationDeliveryIds.add(userNotificationDelivery.getUserNotificationDeliveryId());
-                        }
+													userNotificationDeliveryIds.add(userNotificationDelivery.getUserNotificationDeliveryId());
+												}
 											%>
 
 												<td class="lfr-<%= userNotificationDeliveryType.getName() %>-column">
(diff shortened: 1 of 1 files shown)


tbigby-kristin:
An example of a non-modifiable UserNotificationDeliveryType class is below. We wish to ensure that users do not turn off their 'email' notification for the Message Boards as we use these to broadcast important messages within our organisation.

```
@Component(
	immediate = true,
	property = { 
			"javax.portlet.name=" + MBPortletKeys.MESSAGE_BOARDS		
	},
	service = UserNotificationDefinition.class
)
public class TBigbyMBAddEntryUserNotificationDefinition extends UserNotificationDefinition {

	public TBigbyMBAddEntryUserNotificationDefinition() {
		super(
			MBPortletKeys.MESSAGE_BOARDS, 0,
			UserNotificationDefinition.NOTIFICATION_TYPE_ADD_ENTRY,
			"receive-a-notification-when-someone-adds-a-new-post-in-a-thread-" +
				"or-category-you-are-subscribed-to");

		addUserNotificationDeliveryType(
			new UserNotificationDeliveryType(
				"email", UserNotificationDeliveryConstants.TYPE_EMAIL, true,
				false));
        }
}
```

sammonsjl:
Hi @tbigby-kristin thank you for the contribution!

/Cc @holatuwol @jonathanmccann @SamZiemer @AngeloJ can one you take a look at this one?

jonathanmccann:
This issue looks to be resolved by [LPS-108901](https://issues.liferay.com/browse/LPS-108901) but if that is not the case please let us know.

sammonsjl:
@jonathanmccann I agree it does appear to be resolved in [LPS-108901](https://issues.liferay.com/browse/LPS-108901).  

I will go ahead and close this pr.  @tbigby-kristin if you feel this hasn't been addressed please don't hesitate to let us know.

sammonsjl:
ci:close

PR #3727, opened 2020-02-28
Lps 109525

SamZiemer:
Hello @robertoDiaz 

https://issues.liferay.com/browse/LPS-109525

This adds an upgrade for Document Library comment subscriptions that were added in LPS-85419.

4e13e23 - This commit is needed in cases when a database has been upgraded and new subscriptions have been added before this upgrade is added. In situations like this, we will never upgrade the old subscriptions as at least 1 subscription with the new className exists. I have added a database backup on LPS-109525 demonstrating this. I want to also note when adding a subscription, we first try to fetch an existing subscription with the same parameters, if one exists, we do not add a new subscription so we do not have a chance of a duplicate value exception without the code I removed

/cc @achaparro


SamZiemer:
ci:test:sf

SamZiemer:
ci:test:relevant

PR #3735, opened 2020-03-05
Lps 109121 2

shuyangzhou:


shuyangzhou:
@brianchandotcom Lily's daily monitoring caught this regression, the original problem was unclear. But it seems 4376ee2 is needed if you want to auto clean up the gradle daemon "on certain platform, for certain people". but doing it in compile is actually slowing thing down, which is defeating the whole purpose of making this daemon changes in the first place. So I am just dropping it for now.
Once peter is back, if he can revisit and figure out why 4376ee2 was needed, we might be able to add it back in a different way.

PR #3749, opened 2020-03-19
LPS-110319 - Cannot add a non-instanciable portlet to a widget page in case it is embedded in a recent webcontent

jkappler:


jkappler:
@ealonso

PR #3769, opened 2020-03-30
LRQA-57288 | master

Sharryxue:
https://issues.liferay.com/browse/LRQA-57288

/ cc @patriciajeanperez
Commit is only for master.


Sharryxue:
Admin error.

PR #3795, opened 2020-04-10
workflowContext should be a unmodifiable map

mcgitty:
The design of workflowContext in KaleoInstance is not fork/thread/cluster safe. There is no database row level locking to prevent dirty reads from and competing writes to workflowContext as a Map or as a string. This also applies to Java code and KaleoTaskInstanceToken.workflowContext and KaleoTimerInstanceToken.workflowContext. Liferay should add explicit declaration of this design and behavior in documentations like https://help.liferay.com/hc/en-us/articles/360028818852-Leveraging-the-Script-Engine-in-Workflow


sammonsjl:
ci:close

PR #3814, opened 2020-04-24
LPS-110927 Update Liferay.Portlet.openWindow utility

markocikos:
Hey @wincent ,

This is a followup on https://github.com/wincent/liferay-portal/pull/219

Changes:
- Added loading indicator to show while iframe is loading.
- Added delegation listener cleanup
- Trimmed `processClose` use
- Renamed `<IframeHTML>` and `<BodyHTML>` to `<Iframe>` and `<Body>`
- Updated tests

A challenge with iframes and React is that re-rendering causes requests to iframe url, which is not acceptable. There is also a layer of difficulty with event delegation, which gets destroyed in re-render. Normally, we could put such delegation on `document.body`, but that cannot be applied here as events within iframe do not propagate out of it.

I tried various ways to prevent re-rendering, or to delay event delegation init to last render, but to no success. In the end, I managed to achieve solution by managing spinner directly, through `ref` and 'hide' class, avoiding re-renders. Default close button (x) in `ClayModal`  does a re-render before calling `onClose`, I believe to start fade out. As this gets in the way, I needed to replace it with a button that just does `processClose`.

Please review the code.

Thank you!

/cc @jbalsas


markocikos:
Oops, wrong remote.

PR #3815, opened 2020-04-28
Use the 'localizable' flag to avoid exploding the XML

mcgitty:
When one field has multi-language values, all fields are duplicated by the number of "available-locales" of the root element. For example, a HTML field with images encoded in "data" URL scheme (RFC-2397) may be language neutral and very large. Duplicating it in all available languages is meaningless, wasteful and bad for performance. It also explode the amount of duplicate data being indexed by the search engine.

The DDM structure already has the 'localizable' flag. If a field is marked as "localizable: false", this change makes JournalConverter not to duplicate the content into all available languages.

### DIFF
--- a/modules/apps/journal/journal-service/src/main/java/com/liferay/journal/internal/util/JournalConverterImpl.java
+++ b/modules/apps/journal/journal-service/src/main/java/com/liferay/journal/internal/util/JournalConverterImpl.java
@@ -509,6 +509,10 @@ public class JournalConverterImpl implements JournalConverter {
 
 		String name = dynamicElementElement.attributeValue("name");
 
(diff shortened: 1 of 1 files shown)


sammonsjl:
@mcgitty thank you very much for the contribution and rebasing it on master for us!

@holatuwol @jonathanmccann @SamZiemer @AngeloJ can one you take a look at this one?

Note: CLA accepted here: https://issues.liferay.com/browse/LPS-68424

sammonsjl:
Please see https://github.com/liferay-echo/liferay-portal/pull/2712 for details.

sammonsjl:
Hi @mcgitty, merged and back ported to 7.3.x here: https://github.com/brianchandotcom/liferay-portal/pull/95355
Thanks!

sammonsjl:
ci:close

PR #3849, opened 2020-06-12
LPS-113663 generate

nhpatt:
Sorry for the delay, I asked frontend infra to see if there is another way of not having to exclude the sf of package.json, not yet. 

Also we can't exclude only index.html (we need the folder) because if the source formatter is called before `clean deploy`, the file won't be there and source formatter will complain because having an exclusion that does not have any file :( (I've restricted it to dist folder)


nhpatt:
ci:close

PR #3891, opened 2020-07-30
LPS-116844 Allow questions to be configured with a root topic

javierdearcos:


javierdearcos:
@nhpatt 

I left the last commit with the things related with configuration to be easier to add the part that is missing

PR #3956, opened 2020-09-23
Commerce 4862 escape address fields

andrea-ale-sbarra:


andrea-ale-sbarra:
the fix was done by reverting last wrong commit see #94492

PR #3957, opened 2020-09-23
LPS-121300 only delete expando tables related to the specific instance where the import is ocurring

cleydyr:


cleydyr:
Sent by mistake.

PR #3968, opened 2020-10-06
LPS-121070 The Multiple IDP selector form has bad width

stian-sigvartsen:
Using inline styles isn't desirable, but I think this is just a short term solution until we replace the saml-hook with a portlet that works alongside the login process much like MFA. Then we would get the standard styling.


stian-sigvartsen:
Wrong repo

PR #4046, opened 2020-12-15
LPS-124870 Use disabled input ID box in click goal target instead of link

dgarciasarai:
**Description**
From a [feature request](https://issues.liferay.com/browse/LPS-118605), we are going to add the ability of selecting any element as a click target along with the OOTB ones (button and links with ID).

**Solution**
- Display a different message inside the Click goal section: "_Select a clickable element to be measured using the select button or by entering its ID in the field below._"
- Change "Set Element" and "Edit Element" button copies to "Select Clickable Element" and "Change Clickable Element"
- Use a ClayInput (disabled for now, but in the next sub-task is to be enable to let the user write its ID) instead of a link to display the selected ID, with a label "Element ID" and a help text (question-circle icon). The input has also a tooltip with the selected target
- Selected target is now displaying without # inside the input. Tooper and Popover in the highlighted clickable elements (buttons or links with ID) remains the same (with #)
- View button (eye button) does the same as the previous link: scroll to the selected target and hightlights
- `goalTarget` is stored in the database with the same format as before (#myId)

**Screenshots**

Screenshot 2020-12-15 at 11 23 45 Screenshot 2020-12-15 at 11 23 59
Screenshot 2020-12-15 at 11 33 11 Screenshot 2020-12-15 at 10 36 27


dgarciasarai:
https://github.com/liferay-frontend/liferay-portal/pull/620.

PR #4118, opened 2021-02-10
LPS-127093 partial revert

stian-sigvartsen:
@martamedio We accidentally sent this change with LPS-127093 . It belongs to LPS-125272


stian-sigvartsen:
Wrong fork

PR #4133, opened 2021-02-25
LPS-128334 Detect when a RuntimeException wraps a PortalException an use the correspondent ExceptionMapper

javierdearcos:


modules/apps/portal-vulcan/portal-vulcan-impl/src/main/java/com/liferay/portal/vulcan/internal/jaxrs/exception/mapper/ExceptionMapper.java
```diff
@@ -19,13 +19,30 @@
...
  */
 public class ExceptionMapper extends BaseExceptionMapper<Exception> {
 
+	@Override
+	public Response toResponse(Exception exception) {
+		Throwable throwable = exception.getCause();
+
+		javax.ws.rs.ext.ExceptionMapper<Throwable> exceptionMapper =
+			_providers.getExceptionMapper(
+				(Class<Throwable>)throwable.getClass());
```
javierdearcos, line 37: Don't like at all this cast but it is the only way that I find to implement working code that passes the Source Formatter

modules/apps/portal-vulcan/portal-vulcan-impl/src/main/java/com/liferay/portal/vulcan/internal/jaxrs/exception/mapper/PrincipalExceptionMapper.java
```diff
@@ -44,7 +44,7 @@ protected Problem getProblem(PrincipalException principalException) {
 			status = Response.Status.NOT_FOUND;
 		}
 
-		return new Problem(status, null);
+		return new Problem(status, principalException.getMessage());
```
javierdearcos, line 47: I'm not sure if we were avoiding sending the exception message here for some reason, but I felt it is more correct to give more information to the API consumer

javierdearcos (commented):
Added some comments to review

PR #4149, opened 2021-03-08
LPS-128744 Fix broken DOM structure caused by LPS-120600

georgel-pop-lr:
LPS-128744 Fix broken DOM structure caused by LPS-120600, DOM contains an empty element p and asset-actions.jsp content places outside his parent element span. Change asset_actions.jsp parent element span with div for markup validity as actions contain a div and span can't be parent of a div.

Resend from frontend review to echo team as per request liferay-frontend#870 (comment)


georgel-pop-lr:
Open it by mistake, closing it.

PR #4158, opened 2021-03-16
LRQA-65879 | master

Sharryxue:
https://issues.liferay.com/browse/LRQA-65879

/cc @patriciajeanperez 
Commit is only for master, thanks.


Sharryxue:
Admin error

PR #4181, opened 2021-04-13
LPS-130554 prevent IndexOutOfBoundsException at upgrade by creating n…

sammso:
…ew element


sammonsjl:
Hi @sammso we can only take fixes against master.  Looking through the code in master it looks like the implementation is completely different now so maybe this is no longer an issue there?  I will close for now but please reopen if I missed anything.

sammonsjl:
ci:close

PR #4182, opened 2021-04-13
LPS-130555 pass classloader to read correct bundle instead of current class

sammso:


sammso:
Realized that this is not good enough

PR #4183, opened 2021-04-13
LPS-130567 Fix new lines not taken into account in form description

georgel-pop-lr:
LPS-130567 Fix new lines not taken into account in paragraph element for form description


georgel-pop-lr:
Closed, as I opened by mistake for this user.

PR #4198, opened 2021-04-27
Use tooltip-support-web as global tooltip provider and remove unneccessary ClayTooltipProvider components

bryceosterhaus:
cc @matuzalemsteles @wincent @markocikos

I was taking a look at tooltips in DXP and I think we can get away with purely using the `tooltip-support-web` module instead of making changes to how the ClayTooltipProvider component works.

1. I added `[title]:not(data-tooltip-disable)` and `[data-title]:not(data-tooltip-disable)` as selectors for our global tooltip.
2. I removed any use of just `<ClayTooltipProvider>` where we didn't need any customization.
3. Some uses were only using it for the `delay` prop, so I just added that as a `data-tooltip-delay` attribute
4. For uses of a custom `contentRenderer`, I replaced those titles with the html version of that custom component by utilizing `ReactDOMServer.renderToString`.

The last part is the one I am must unsure about. Anyone have any concerns or see issues with this?

### DIFF
--- a/modules/apps/commerce/commerce-frontend-js/package.json
+++ b/modules/apps/commerce/commerce-frontend-js/package.json
@@ -20,11 +20,11 @@
 		"@clayui/shared": "3.5.1",
 		"@clayui/sticker": "3.3.0",
 		"@clayui/table": "3.1.0",
-		"@clayui/tooltip": "3.4.5",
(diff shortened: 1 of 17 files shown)


bryceosterhaus:
wrong repo

PR #4224, opened 2021-05-19
LPS-132444 speed up lar import with a ThreadLocal cache

mmariuzzo:
Add a ThreadLocal cache to speed up imports of JournalArticle and DLFileEntry.
Introduced a method to show import progress also on log (every 2%) to help to guess how long it will take


mmariuzzo:
I'm closing it because the contributed code has been included in wider change

PR #4340, opened 2021-08-19
LPS-137532 Have search container return only the orphan portlets to be displayed

noornajj:
https://issues.liferay.com/browse/LPS-137532

The issue here was that the orphan widgets page would not update with the correct number of entries when the number of entries per paged was less than the total number of entries.
To fix this, I made sure the results in the orphanPortletsSearchContainer only holds the portlets to be displayed instead of all the portlets.


noornajj:
ci:test:sf

noornajj:
ci:test:relevant

PR #4364, opened 2021-09-08
master_LPS-138778

davidtellolr:
Hi team,

This PR is a fix of the regression bug produced by LPS-94001 (commit 3967db8fd172cca15b6e5cf3144c0cab4702d20f) in the rename Display Page Template action.

With the change made in LPS-94001, when a Display Page Template is renamed, only the draft Layout is changed but not the main Layout. Thereby producing that when the Display Page Template name must be shown, the old value is used.

To fix the error, I added the necessary code to save the main layout.

Please, let me know if you have any doubt.


davidtellolr:
ci:test:sf

PR #4373, opened 2021-09-14
LPS-139184

mbowerman:
Comments from @javalosjr96:

> @mbowerman
The permission check for adding the path to the table view was removed from view_entries in Journal-Web in order to fix the skewed row that is created by the varying differences in permissions. The permission check was added to JournalHelperImpl in the getAbsoultePath method. If a user does not have permission to view a certain folder, the folder's name is replaced with ". . ." as indicated by the PT member in the [PTR ticket](https://issues.liferay.com/browse/PTR-2643) .

/cc @javalosjr96 

https://issues.liferay.com/browse/LPS-139184


mbowerman:
ci:test:sf

mbowerman:
ci:test:r

PR #4408, opened 2021-10-08
[TS] LPS-140390 | master

samanzanedo:
**Description**
There are some situations in which database indexes are created using a limited length in their column(s) definition.

This is done, for example, when a table column is defined as LOB but an index is considered as necessary based on that column. Thus, only a 'partial' character content from that column is used.

There is available the following property for configuring a limit in the length of index column definition:
`database.string.index.max.length `
This property is defined based on the target database using an internal suffix. Currently,[ it can be defined for Mysql and MariaDB databases](https://github.com/liferay/liferay-portal/blob/master/portal-impl/src/portal.properties#L1719#L1726):
```
#
# Set the max length for string type of index.
#
# Env: LIFERAY_DATABASE_PERIOD_STRING_PERIOD_INDEX_PERIOD_MAX_PERIOD_LENGTH_OPENBRACKET_MARIADB_CLOSEBRACKET_
# Env: LIFERAY_DATABASE_PERIOD_STRING_PERIOD_INDEX_PERIOD_MAX_PERIOD_LENGTH_OPENBRACKET_MYSQL_CLOSEBRACKET_
#
database.string.index.max.length[mariadb]=255
database.string.index.max.length[mysql]=255
```
	
If you try to use this property in an Oracle database:
`database.string.index.max.length[oracle]=255`

the transformer task that handles the SQL syntax for the index will fail because of it is not a valida syntax for Oracle (_ORA-00904: invalid identifier error code_).

Furthermore, there could be situations that creating an index (in an Oracle database environment) cause another error because of the size limitation under a default configuration of database block size (8 KB)
(_ORA-01450: Maximum Key Length (6398) error code_)

**Proposed solution**
So the proposed solution is to increase the handling of that transformer task to allow using Oracle indexes syntax. 
The strategy is to define a SQL syntax with a function-based index, using the `substr()` function in conjunction with the value defined in the property. 

**References**
https://issues.liferay.com/browse/LPS-140390
https://issues.liferay.com/browse/PTR-1176

### DIFF
--- a/portal-impl/src/com/liferay/portal/dao/db/BaseDB.java
+++ b/portal-impl/src/com/liferay/portal/dao/db/BaseDB.java
@@ -861,6 +861,20 @@ public abstract class BaseDB implements DB {
 				new Filter(dbType.getName())),
 			-1);
 
+		String dbName = dbType.getName();
+
+		if (dbName.equals(DBType.ORACLE.toString())) {
+			return _applyOracleMaxStringIndexLengthLimitation(
+				template, stringIndexMaxLength);
+		}
+
+		return _applyMySQLMaxStringIndexLengthLimitation(
+			template, stringIndexMaxLength);
+	}
+
+	private String _applyMySQLMaxStringIndexLengthLimitation(
+		String template, int stringIndexMaxLength) {
+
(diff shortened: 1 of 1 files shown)


samanzanedo:
ci:test:relevant

PR #4476, opened 2021-11-29
Using JWT assertion as Client Authentication method for OIDC

arthurchan35:
Hey @stian-sigvartsen 
Only the last 3 commits are for this task.

We can discuss about it some time later.


modules/apps/portal-security-sso/portal-security-sso-openid-connect-impl/src/main/java/com/liferay/portal/security/sso/openid/connect/internal/util/OpenIdConnectTokenRequestUtil.java
```diff
@@ -101,12 +101,20 @@ private static OIDCTokens _requestOIDCTokens(
 
 		URI uri = oidcProviderMetadata.getTokenEndpointURI();
 
-		ClientID clientID = new ClientID(openIdConnectProvider.getClientId());
-		Secret secret = new Secret(openIdConnectProvider.getClientSecret());
```
stian-sigvartsen, line 105: @arthurchan35 is it ok to couple `OpenIdConnectTokenRequestUtil` to our `OpenIdConnectProviderImpl`?  The instance is obtained from `OpenIdConnectProviderRegistry` which is a SPI.

We actually used to have an unsafe cast to `OpenIdConnectProviderImpl` , but this was removed with our recent changes for the refresh token impl.

stian-sigvartsen:
This PR covers a lot of ground, which is great! I can see it being merged as multiple tickets. The key thing as I see it is making the UX good for adding the private key, and describing the feature options using terminology familiar to the admins (maybe you've already checked this). Looking forward to exploring it tomorrow.

PR #4640, opened 2022-03-03
LPS-131529 Using original names when downloading files

jan-tosovsky-cz:
Prefer using original names instead those parsed form download URL, see details in https://issues.liferay.com/browse/LPS-131529


jan-tosovsky-cz:
This was fixed differently by LR staff so closing without merging.

PR #4766, opened 2022-05-10
[TS] LPS-153234 If categoryIds is empty in isMissingRequiredCategory return false

evibudai:
Hi Team,

https://issues.liferay.com/browse/LPS-153234

**Issue**:
Draft saving fails in Blogs widget when adding a "New entry" if we have a vocabulary with associated asset type "Blogs Entry" set to "Required".
With every draft saving we run isMissingRequiredCategory() which fails, because the "categoryIds" array is empty and we try to check the contents of it.
I checked whether categoryIds should be empty or not: it comes from ServiceContext.getAssetCategoryIds() which is always empty in this scenario.

**Solution**:
I added a validation in isMissingRequiredCategory to check if `long[] categoryIds` is empty. If it's empty we return with false. The draft saving is succeful.

image

Could you please review it?
Thank you!


evibudai:
I'm closing this because of the bad base repository

PR #4777, opened 2022-05-18
LPS-153731 Hidden pages are not displayed in the layout selector form field

matthiasblaesing:
https://issues.liferay.com/browse/LPS-153731


matthiasblaesing:
Fix is already provided by https://github.com/liferay-echo/liferay-portal/pull/8241 and integrated as 71fd574cb372034e41ab92bca3ab808272981f0e

PR #4823, opened 2022-06-17
Lps x publication cache first

tinatian:


modules/apps/change-tracking/change-tracking-service/src/main/java/com/liferay/change/tracking/service/impl/CTEntryLocalServiceImpl.java
```diff
@@ -226,6 +226,20 @@ public boolean hasCTEntries(long ctCollectionId, long modelClassNameId) {
 		return true;
 	}
 
+	@Override
+	public boolean hasCTEntry(
+		long ctCollectionId, long modelClassNameId, long modelClassPK) {
+
+		int count = ctEntryPersistence.countByNotC_MCNI_MCPK(
```
david-truong: should be countByC_MCNI_MCPK

PR #4897, opened 2022-07-27
LPS-159086 should not use compileOnly for gson lib

simonjhy:


simonjhy:
ci:test:gradle-plugins

PR #4962, opened 2022-09-06
LPS-162225 Make it clear that we should do the startsWith check first.

mbowerman:
startsWith is O(n) with respect to the length of the _argument_ string.
In our case, the input string is "data:image/" which is always exactly
11 characters long.

contains is O(n) with respect to the length of the String being operated
on. It is not uncommon for this number to be in the millions for Base64
Strings.

Therefore for these Base64 Strings there is a noticeable
difference in performance if we do startsWith first and return early,
compared to if we do contains first and then startsWith.


mbowerman:
ci:test:sf

PR #5016, opened 2022-10-09
LPS-156658: Synchronize permissions of DLFileEntryType and corresponding DDMStructure

matthiasblaesing:
Each DLFileEntryType is tied to a DDM structure, that holds the metadata definitions defined in the document type itself.

The upload forms (and most probably other code) checks the permissions on the DDM structure to determine if the fields may be viewed/entered by a user. So the permissions on the structure need to be sychronized with the DLFilEntryType.


adolfopa:
The approach looks good @matthiasblaesing. Can you send next prs to @liferay-lima?

matthiasblaesing:
@adolfopa thank you for the hint. I'll update the work and open a new PR against the referenced fork.

PR #5033, opened 2022-10-13
LPS-165316 Keep the Experience Selector Open after Creating a New Experience

beltranrengifo:
## Motivation

Improve the UX of the Experiences dropdown.
In this PR we keep the panel open on Experience creation: being precise, we reopen the dropdown on modal close.
Also now we keep the panel open on experience deletion.

## Solution proposed

We get rid of the blur/focus events we used in the button and the dropdown, as they had conflicts between them. It was working but not in a very smooth way: the button blur event was closing the panel, so a new focus event was implemented in the panel reopening it, this allowed the user to click in the button and then in the panel without losing it.

Now we use a `clickoutside` hook with a ref to the dropdown. This way the panel only gets closed if it's open and the user clicks away.
In addition we force the reopening of the dropdown on modal close.

## How to test
- navigate to home page
- edit the page
- click on experiences selector and open the dropdown
- create a new experience
- the dropdown should be open on modal close
- delete an experience
- the dropdown should be open
- click outside the dropdown
- the dropdown should be closed


beltranrengifo:
ci:close

PR #5204, opened 2023-01-13
LPS-172719 Remove Dom4jUtil usage from tld-formatter

julschong:


util-taglib/src/META-INF/liferay-theme.tld
```diff
@@ -6,7 +6,7 @@
 	<jsp-version>1.1</jsp-version>
 	<short-name>liferay-theme</short-name>
 	<uri>http://liferay.com/tld/theme</uri>
-	<description>Provides the Liferay Theme component tags, prefixed with <![CDATA[<code>theme:</code>]]>.</description>
```
julschong: @tinatian Hi Tina,
I tried the same way we changed Service Builder on TLD Formatter
This output removed the space before and after `<![CDATA[....]]>`

this is because we outputFormat.setPadedText(false) 
The original Dom4jUtil.toString manually got rid of those padded text without affect these 2 spaces.

Is this result acceptable?
otherwise, we may need to go back to the original manual space removal logic

julschong: After research, I don't think it would acceptable since those spaces are important in the sentence
image

PR #5209, opened 2023-01-16
COMMERCE-10535 Enable customised cookie configuration modal for product comparison

dsanz:
Hi @brianchandotcom Bryce and myself reviewed the language keys as per your [request](https://github.com/brianchandotcom/liferay-portal/pull/128545#issuecomment-1379563146). Other than that, I've just made minor changes because of the rebase from master. Thank you

### References

- https://issues.liferay.com/browse/COMMERCE-10535
- This PR is forwarded from https://github.com/liferay-frontend/liferay-portal/pull/2938 which has non-related test failures, verified both from Commerce and FI QA

### What is the goal of this PR?
To make product comparison feature to obey user consent for functional cookies. This feature requires a non-essential cookie which must be accepted by users.

In doing so, we added a DXP JS api to handle cases like this in the future, so that other applications don't need to reinvent the wheel when some feature requires cookie acceptance

Before this change, it was possible to select products for comparison, and the minicompare component was rendered. Also, the `COMMERCE_COMPARE_cpDefinitionIds_*` cookie was created and stored in the user's device.

After this change, the product selection process ensures functional cookies are accepted before setting `COMMERCE_COMPARE_cpDefinitionIds_*` cookie. Also, product comparison widget is not displayed until user these cookies are accepted. Cookie consent modal is shown to allow user to give consent.


dsanz:
wrong destination...

PR #5249, opened 2023-02-12
Password History should check last n passwords

BurhanNoori:
When enabling the password history with 2, the last 2 passwords of the user must be checked. For eg if user has 1st password apple, then it changes to ball and then it changes to cat. He should not allow to set apple. A same password validation restrict him to set "cat" again so the last two passwords would be ball and apple excluding the current one(not including the current password).


sammonsjl:
Hi @BurhanNoori, thank you for your contribution!  Would you please send it to Brian Chan directly: https://github.com/brianchandotcom/liferay-portal/pulls so that we can make sure it gets included.

PR #5276, opened 2023-02-27
LPS-176789 Migrate DateTimeInput to a functional Component + TypeScript

p2kmgcl:
Motivation
=======
Before adding the new "range" property to the component, I have updated it to be a functional component and migrated the code to typescript.

https://issues.liferay.com/browse/LPS-176789

Proposed Solution
========

Using functional components allow us write modern React code that increases re-usability and simplicity. For example, I've been able to reuse the `usePrevious` hook instead maintaining two different states by myself.

Also migrating it to TypeScript is part of portal FrontEnd goals (ex. Clay is written in TypeScript). It enables a high-quality development experience and increases robustness with type checks. In the case of React components, it also allow us to stop depending on the `prop-types` library, having prop checking during the deploy process and zero cost in runtime.

Both `tsconfig.json` and `*.d.ts` files are auto-generated and updating during deploy, but they must be committed.

There are no UI or behavior changes, the code should work the same way.

Checklist (optional):
========
- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] Any dependent changes have been merged and published in downstream modules


p2kmgcl:
ci:test:sf

p2kmgcl:
ci:test:relevant

PR #5376, opened 2023-04-12
LPS-180803 Fix NPE on null values

EmTeedee:
The column is default null, so that should be handled.


sammonsjl:
Hi @EmTeedee thank you for your contribution!  Would you please send this pull request directly to Brian Chan at: https://github.com/brianchandotcom/liferay-portal/ that way we can make sure to get it included as soon as possible.

EmTeedee:
ci:forward

PR #5398, opened 2023-04-27
LPS-182073 Add a new type of client extension called Frontend Dataset Cell Renderer

dsanz:
In this PR we are aligining the implemented UI to the designs. This UI belongs to the Client Extensions, in particular, when creating a new type of frontend extensions called "Frontend Dataset Cell Renderer".

Alignment basically consists of adding the text message as the rest of the UI already conformed with the expected outcome, including naming. After some discussion, we agreed that the user-friendly name of this extension is "Frontend Dataset Cell Renderer" as it perfectly reflects its intent. Internally, we use the `FDSCellRenderer` nomenclature which permeates to Java, JSP and language key naming.

This is a followup of https://github.com/liferay-frontend/liferay-portal/pull/3253


dsanz:
ci:test:sf

PR #5428, opened 2023-05-11
LPS-179688 As an Instance Administrator, I want to be able to configure that if the users have the option to give their password at registration

thabet91:


brianchandotcom:
@thabet91 this one is merged

PR #5469, opened 2023-06-05
Draft PR of accessibility menu.

ethib137:
This is just an initial draft, but any reviews would be helpful.

All comments and suggestions are appreciated.

https://issues.liferay.com/browse/LPS-178191

@john-co , @matuzalemsteles , @marcoscv-work , @pat270


modules/apps/accessibility/accessibility-menu-web/src/main/resources/META-INF/resources/js/AccessibilitySetting.js
```diff
@@ -0,0 +1,93 @@
...
+
+import {SETTINGS_STRING_VALUES_MAP, getSettingValue} from './util';
+
+const KEY_EVENT = 'Enter';
+
+const AccessibilitySetting = ({setting}) => {
+	const {className, defaultValue, key, label, sessionClicksValue} = setting;
+
+	const [value, setValue] = useState(
+		SETTINGS_STRING_VALUES_MAP[sessionClicksValue]
```
matuzalemsteles: You can get lazy initialization here too instead of creating a `useEffect`.

```jsx
const [value, setValue] = useState(() => getSettingValue(defaultValue, sessionClicksValue, key));
```

modules/apps/accessibility/accessibility-menu-web/src/main/resources/META-INF/resources/js/AccessibilitySetting.js
```diff
@@ -0,0 +1,93 @@
...
+
+	return (
+		<div className="c-mb-3">
+			<ClayToggle
+				disabled={disabled}
+				label={label}
+				onKeyDown={(event) => {
+					if (event.key === KEY_EVENT) {
+						handleToggle(!value);
+					}
```
matuzalemsteles: Maybe we can add this directly into Clay, pressing `Enter` should call the `onToggle` callback.

modules/apps/accessibility/accessibility-menu-web/src/main/resources/META-INF/resources/js/AccessibilityMenu.js
```diff
@@ -0,0 +1,85 @@
...
+					getSettingValue(
+						setting.defaultValue,
+						setting.sessionClicksValue,
+						setting.key
+					)
+				);
+		});
+
+		return () => {
+			Liferay.detach(OPEN_ACCESSIBILITY_MENU_EVENT_NAME, onOpenChange);
```
matuzalemsteles: I could be wrong but I think `Liferay.detach` will just remove the listener based on the function reference too, as above you create an anonymous function it's a different reference to `onOpenChange` for security I think the best would be to create a named function inside the `useEffect` and pass it to the `Liferay.on` and `Liferay.detach` method.

modules/apps/accessibility/accessibility-menu-web/src/main/resources/META-INF/resources/js/util.js
```diff
@@ -0,0 +1,42 @@
...
+export const SETTINGS_STRING_VALUES_MAP = {
+	false: false,
+	true: true,
+	undefined,
+};
+
+export function getSettingValue(defaultValue, sessionClicksValue, key) {
+	if (
+		!themeDisplay.isSignedIn() ||
+		SETTINGS_STRING_VALUES_MAP[sessionClicksValue] === undefined
```
matuzalemsteles: Wouldn't it be better to just check `sessionClicksValue === undefined` or `typeof sessionClicksValue === 'undefined'`? What would be the purpose of `SETTINGS_STRING_VALUES_MAP`?

modules/apps/accessibility/accessibility-menu-web/src/main/resources/META-INF/resources/js/util.js
```diff
@@ -0,0 +1,42 @@
...
+			key,
+			localStorage.TYPES.PERSONALIZATION
+		);
+
+		if (localStorageValue in SETTINGS_STRING_VALUES_MAP) {
+			return SETTINGS_STRING_VALUES_MAP[localStorageValue];
+		}
+		else {
+			return SETTINGS_STRING_VALUES_MAP[defaultValue];
+		}
```
matuzalemsteles: I think just checking `localStorageValue !== null` would be enough to not need `SETTINGS_STRING_VALUES_MAP`.

And we could just return the value `defaultValue` or `localStorageValue` instead of using the constant for that `SETTINGS_STRING_VALUES_MAP`.

matuzalemsteles (commented):
@ethib137 it looks good, I liked that some things are kind of generic on the client side, so we can grow the options menu over time, but I think we should implement something that doesn't just depend on changing the CSS class because for reduced motion to work for all cases of usage we need something implemented in the global state to disable animations in JS https://clayui.com/docs/components/reduced-motion.html#js.

I think we can easily get around this by changing the state of `useState` in the `AccessibilitySetting` component to `useLiferayState` so we can have state readers in other places.

john-co:
ci:test:bundle

john-co:
Would you be able to send this PR to liferay-peds so we can leverage PR tester? `ci:test:bundle`, `ci:test:sf`, etc.

modules/apps/accessibility/accessibility-menu-web/src/main/resources/META-INF/resources/js/AccessibilitySetting.js
ethib137: Well, since the toggle is a form element, pressing enter when focused and inside a form should submit the form. It's only in this case where it's not in a form and I want to immediately save the setting that it makes sense to toggle on 'Enter'.

modules/apps/accessibility/accessibility-menu-web/src/main/resources/META-INF/resources/js/AccessibilityMenu.js
ethib137: Thanks @matuzalemsteles! That's what I meant to do. Good catch! 😄

modules/apps/accessibility/accessibility-menu-web/src/main/java/com/liferay/accessibility/menu/web/internal/template/AccessibilityTemplateContextContributor.java
```diff
@@ -0,0 +1,85 @@
...
+				AccessibilitySettingsUtil.getAccessibilitySettings(
+					httpServletRequest)) {
+
+			if (_isSettingEnabled(accessibilitySetting)) {
+				sb.append(StringPool.SPACE);
+				sb.append(accessibilitySetting.getClassName());
+			}
+		}
+	}
+
```
rafaprax, line 67: Hey @ethib137, In my point of view, this method should be part of the `AccessibilitySetting` implementation(the `AccessibilitySetting` impl should know when it is enabled or not).

My suggestion here is to move the `_isSettingEnabled` to the `AccessibilitySetting` class and name this method `isEnabled`. 

Another thing is to use the `GetterUtil.getBoolean` instead of the `Boolean.getBoolean`.

Thx

modules/apps/accessibility/accessibility-menu-web/src/main/java/com/liferay/accessibility/menu/web/internal/template/AccessibilityTemplateContextContributor.java
```diff
@@ -0,0 +1,85 @@
...
+	private boolean _isSettingEnabled(
+		AccessibilitySetting accessibilitySetting) {
+
+		if (accessibilitySetting.getSessionClicksValue(
+			).equals(
+				"undefined"
+			)) {
+
```
rafaprax: Hey @ethib137, avoid the chained calls in the if expression. Thx

modules/apps/accessibility/accessibility-menu-web/src/main/java/com/liferay/accessibility/menu/web/internal/display/context/AccessibilitySetting.java
```diff
@@ -0,0 +1,99 @@
...
+	public void setSessionClicksValue(String sessionClicksValue) {
+		this.sessionClicksValue = sessionClicksValue;
+	}
+
+	@JsonProperty
+	protected String className;
+
+	@JsonProperty
+	protected String defaultValue;
+
```
rafaprax: Hey @ethib137, according to the usages of the `defaultValue` it should be a boolean. Could please use the boolean type here?

Thx

modules/apps/accessibility/accessibility-menu-web/src/main/java/com/liferay/accessibility/menu/web/internal/display/context/AccessibilitySetting.java
```diff
@@ -0,0 +1,99 @@
...
+import java.util.function.Consumer;
+
+/**
+ * @author Evan Thibodeau
+ */
+@JsonInclude(JsonInclude.Include.NON_DEFAULT)
+public class AccessibilitySetting {
+
+	public static AccessibilitySetting create(
+		String label,
```
rafaprax: Hey @ethib137, I think using a constructor is a better fit for your use case for now. If the building of an AccessibilitySetting becomes complex we may apply the Builder pattern here. So, for now, let's go with just a constructor.

The constructor should look like this one. 

```

public AccessibilitySetting(String className, boolean defaultValue, String label, String key,
 String sessionClicksValue) {
      // Initilize variables here
      ...
}
```

Thx

modules/apps/accessibility/accessibility-menu-web/src/main/java/com/liferay/accessibility/menu/web/internal/display/context/AccessibilityMenuDisplayContext.java
```diff
@@ -0,0 +1,47 @@
...
+
+	public AccessibilityMenuDisplayContext(
+		HttpServletRequest httpServletRequest) {
+
+		_httpServletRequest = httpServletRequest;
+	}
+
+	public List<Object> getAccessibilitySettings() {
+		ObjectMapper objectMapper = new ObjectMapper();
+
```
rafaprax: Hey @ethib137, please use the `JSONUtil.toJSONArray` instead of the jackson ObjectMapper. For your use case using the JSONUtil is enough. In the end, your code will look like this:
```
public JSONArray getAccessibilitySettingsJSONArray() {
      return JSONUtil.toJSONArray(
           AccessibilitySettingsUtil.getAccessibilitySettings(_httpServletRequest),
           accessibilitySetting -> {
               //blablabla
           }
      )
}
```

Here is an example https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-data-set/frontend-data-set-views-web/src/main/java/com/liferay/frontend/data/set/views/web/internal/display/context/FDSViewsDisplayContext.java#L52-L66

rafaprax: With this change, you'll be able to remove the jackson annotations from the AccessibilitySetting class as well.

rafaprax (changes requested):

modules/apps/accessibility/accessibility-menu-web/src/main/java/com/liferay/accessibility/menu/web/internal/display/context/AccessibilitySetting.java
ethib137: Thanks @rafaprax , I need to make it a boxed Boolean because I need to have a value that indicates that sessionClicksValue has not been set by the user. Boxed boolean will let me use null for that. We need this because on the frontend if sessionClicksValue has not been set we will try to get the localStorage value, then if that's not set we will set it to the defaultValue. 

Let me know if you think there is a better way to handle this.

modules/apps/accessibility/accessibility-menu-web/src/main/java/com/liferay/accessibility/menu/web/internal/template/AccessibilityTemplateContextContributor.java
ethib137: Makes sense! Thanks!

modules/apps/accessibility/accessibility-menu-web/src/main/java/com/liferay/accessibility/menu/web/internal/display/context/AccessibilityMenuDisplayContext.java
ethib137: Sounds good.. thanks.

modules/apps/accessibility/accessibility-menu-web/src/main/java/com/liferay/accessibility/menu/web/internal/display/context/AccessibilitySetting.java
ethib137: Makes sense.

modules/apps/accessibility/accessibility-menu-web/src/main/resources/META-INF/resources/js/util.js
ethib137: Yep, I was able to rework this to get rid of SETTINGS_STRING_VALUES_MAP. Initially I found it complex because sessionClicks only returns a string, but I'm able to process that on the backend and make it simpler on the frontend for us. Thanks.

modules/apps/accessibility/accessibility-menu-web/src/main/resources/META-INF/resources/js/util.js
ethib137: Same as above.

modules/apps/accessibility/accessibility-menu-web/src/main/java/com/liferay/accessibility/menu/web/internal/template/AccessibilityTemplateContextContributor.java
ethib137: > AccessibilitySetting impl should know when it is enabled or not

@rafaprax does this mean I should have an interface and impl for AccessibilitySetting? If so, what is the deciding criteria we use for having an interface and impl over just a class? Thanks!

modules/apps/accessibility/accessibility-menu-web/src/main/java/com/liferay/accessibility/menu/web/internal/template/AccessibilityTemplateContextContributor.java
ethib137: Sounds good. Thanks!

ethib137:
@matuzalemsteles for reduced-motion you recommend the following:
> I think we can easily get around this by changing the state of useState in the AccessibilitySetting component to useLiferayState so we can have state readers in other places.

I'm going to remove reduced-motion from this final pull, and we will add it in a different story, but I'm curious how you would have this work? 

- What value would we need to put in useLiferayState? Would it be all the accessibilitySettingValues? 
- Where would a reader for the reduced-motion value be? What would it do with the value?

modules/apps/accessibility/accessibility-menu-web/src/main/java/com/liferay/accessibility/menu/web/internal/template/AccessibilityTemplateContextContributor.java
rafaprax: Hey @ethib137, the current AccessibilitySetting class is enough. When I mentioned 'impl' I was referring to the current class implementation.

modules/apps/accessibility/accessibility-menu-web/src/main/resources/META-INF/resources/js/AccessibilitySetting.js
matuzalemsteles: Yeah it makes sense to me

matuzalemsteles:
>I'm going to remove reduced-motion from this final pull, and we will add it in a different story, but I'm curious how you would have this work?

Yeah, it makes sense to move this to another task. Basically we will use the [Liferay global state API](https://github.com/liferay/liferay-frontend-projects/blob/master/guidelines/dxp/react_state_management.md#liferays-global-state-api), by using `useLiferayState` instead of `useState` it means that any developer who has access to the atom can read or subscribe to the state of the atom, this allows for example in the taglibs of `react:component` we can add the [`ClayProvider`](https://clayui.com/docs/components/reduced-motion.html#js) component and read the state of the reduced motion atom and set the [`reducedMotion`](https://clayui.com/docs/components/reduced-motion.html#js) property of the context, this will propagate the setting to all Clay components.

Basically this will work reactively once we have the global state so we don't need to implement our own mechanism for that.

marcoscv-work:
Hey guys, can't force the reduce motion only by CSS? I mean, I know it does nor make sense JS handling states for animations when we activate reduce-motion.. but if at the end of the day it works fine just by CSS, for me it is ok :-P

matuzalemsteles (approved):
The JS part looks good to me.

PR #5536, opened 2023-07-10
LPS-189995 riggering un update action on a object, the saving gives an error when attachment field is empty.

gergelyszalay:


gergelyszalay:
Srry, i hit enter by mistake

PR #5552, opened 2023-07-19
LPS-190951

olafk:
A trivial change to implement https://liferay.atlassian.net/browse/LPS-190951 

I'm rarely pullrequesting to portal, so when I do, I'm most likely taking the wrong turn - I hope that the liferay-repository is monitored for pull requests, otherwise I'll keep an eye on this and figure out where better to put it in a while.


olafk:
When I opened the pullrequest, this account was part of the Liferay organization. I believe that's enough. 
Somebody decide if this gets accepted or rejected due to this no longer the case, and/or if you like having the ancient pre-placeholder HTML markup in the product.

PR #5631, opened 2023-09-06
LPS-190851 Restructure filters selection

chestofwonders:
Jira Issue: https://liferay.atlassian.net/browse/LPS-190851

Use cases:

- Create a filter by type:
image

-Fields should be filtered by filter type (date range or string fields:
image
image

- If Selection type is selected and there are no picklists created, a message should appear and there is no possibility to continue:
image

-If there are picklists created, you should be able to show available fields to filter by. In case that a field is being used in another filter, a "in use" pill should appear.
image

- If an already in use field is selected, there should appear a warning message, disabling to continue or save.
image

From user's point of view, the behavior should remains the same as the previous filter creation flow.

**Decisions:**

- References to documentation have been removed from scope
- To build the new custom fields dropdown, I used a DropDownWithItems as in modules/apps/frontend-data-set/frontend-data-set-views-web/src/main/resources/META-INF/resources/js/fds_view/Fields.tsx
The reason behind is to be able to match the items list width with the parent input trigger, because default Clay dropdown components have a fixed width.
- As the current issue is related with the modal window, the filters table itself hasn't been modified, so we need to change the filter's type in the future
image


markocikos:
@chestofwonders please reopen PR on https://github.com/liferay-frontend

markocikos:
ci:close

PR #5857, opened 2024-02-22
LPD-17892

JavierMoral:
Motivation
=======
Solve https://liferay.atlassian.net/browse/LPD-17892
When using property `layout.manage.pages.initial.children=-1`, the layout trees should not use pagination and show all the children of a given node. See property description:
```html
# Set the initial number of child pages to display in the Manage Pages tree.
# Set this to -1 to show all.
layout.manage.pages.initial.children=20
```
This was not properly managed in our code in several uses of the layouts tree.

Steps to Verify
========
1. Add this to portal-ext.properties: layout.manage.pages.initial.children=-1
1. Start up Liferay, log in as Admin
1. Go to Site > Site Builder > Pages
1. Create a few static pages (they can be empty)
1. Create some child pages as well
1. Enable Local Live staging
1. Create a new publication process
1. Scroll down to the Pages section and behold the page tree under the "Pages to Publish" heading
1. All the top level pages are visible in the page tree correctly, however, at the bottom there is a link to "Load More Results"
1. "Load More Results" link is visible under Child pages as well
1. Click Load More Results

Results of Testing

**Expected Result:** This link should not be there at all because all the top level pages are also listed in the page tree because the portal ext entry mentioned in the first point is used. If the value is set to -1, the link to "Load More Results" should not be shown at all because all the pages of the page tree have already been shown and no pages are remaining to be loaded in the page tree.

**Actual Results:** Clicking on "Load More Results" the complete page hierarchy is duplicated and all the pages are now shown twice. The "Load More Results" page is still available after each click on it, resulting into an infinite duplication of the page tree/hierarchy if user keeps on clicking the link. Same behavior is seen with child pages too.

**Another uses of this layouts tree can be tested using the layouts created in these reproduction steps such as:**
1. Clicking in "Page tree" in the product menu will render the layouts tree
2. Creating a page using the "Link to a page of this site" template and selecting the page to link to
3. Creating a structure with a "Link to page" field. Creating a content using this structure will show a button that will render this layout tree

### DIFF
--- a/modules/apps/layout/layout-admin-web/src/main/java/com/liferay/layout/admin/web/internal/util/LayoutsTreeImpl.java
+++ b/modules/apps/layout/layout-admin-web/src/main/java/com/liferay/layout/admin/web/internal/util/LayoutsTreeImpl.java
@@ -367,7 +367,7 @@ public class LayoutsTreeImpl implements LayoutsTree {
 			httpServletRequest, "paginate", true);
 
 		if (paginate &&
-			(PropsValues.LAYOUT_MANAGE_PAGES_INITIAL_CHILDREN > -1)) {
+			(PropsValues.LAYOUT_MANAGE_PAGES_INITIAL_CHILDREN > 0)) {
 
 			return true;
 		}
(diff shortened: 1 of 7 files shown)


JavierMoral:
ci:close

PR #5861, opened 2024-02-23
 LPP-52548 Add creatorId to QuestionFilter.es.js to be able to filter to 1 creator

jgarcialr:


modules/apps/questions/questions-web/src/main/resources/META-INF/resources/js/pages/questions/components/QuestionFilter.es.js
```diff
@@ -132,10 +133,22 @@ const getFilterValues = (form, tags) => {
 		} (keywords/any(x:${filterKeyword}))`;
 	}
 
+	query.creatorId = form.creatorId;
+
+	if (query.creatorId) {
+		const creatorIdFilter = `creatorId eq ${query.creatorId}`;
+
+		query.filterBy =
+			query.filterBy === void 0
```
kevenleone: I never seem this `void 0` applied at portal before, very interesting but I would use something like:

At least is the common for our reality

```suggestion
			query.filterBy
```

The rest looks very good!

kevenleone (commented):
Looks good for me, just this void 0 thing

PR #5904, opened 2024-04-16
LPD-22600 Add DEFAULT_LANGUAGE_ID to indexing

gergelyszalay:
Issue is https://liferay.atlassian.net/browse/LPD-22600. If iuploaded a document, after I change the site dafult language, the search does not displaying the title and content for document.

Rootcause: DLFileEntryModelSummaryContributor.java the getSummary() method cannot get content and title from the document, because the document doesnt have Field.DEFAULT_LANGUAGE_ID and it uses the portal default, and not the document one.

Solution. add to the document indexing the default language id in DLFileEntryModelDocumentContributor.java, so after re-indexing it, it changes the default language together with the localized title and content.

tested it locally and solves the issue


gergelyszalay:
Sorry, wrong repo

PR #5909, opened 2024-04-25
LPD-24098 Fix a11y error with autocomplete and checkboxes

oliv-yu:
https://liferay.atlassian.net/browse/LPD-24098 - Accessibility issue with checkboxes and attribute 'autocomplete'

This [w3c rule](https://rocketvalidator.com/html-validation/attribute-autocomplete-is-only-allowed-when-the-input-type-is-color-date-datetime-local-email-hidden-month-number-password-range-search-tel-text-time-url-or-week) states that `autocomplete` is not a valid attribute for checkbox inputs, but originally `autocomplete=off` was added to address [this bug](https://liferay.atlassian.net/browse/LPS-168259). I tried it out on my Firefox 125.0.2, but I'm not able to reproduce that bug anymore.

I didn't want to just remove the attribute altogether, but the related Stack Overflow solution mentions that another way would be to add it onto the `form` instead: https://stackoverflow.com/questions/5985839/bug-with-firefox-disabled-attribute-of-input-not-resetting-when-refreshing

A few weeks ago, there was another [a11y error](https://liferay.atlassian.net/browse/LPD-20841) about having `autocomplete` for `hidden` inputs, so this change to add it to the encompassing `form` might address some concerns we may have had about regressions there too. 👍 

Let me know if this is fine, thanks for taking a look!


oliv-yu:
https://github.com/liferay-search/liferay-portal/pull/1321

PR #5999, opened 2024-07-22
LPD-31812 Remove LPD-10754 Feature Flag

dsanz:
In this PR we are removing all usages of `LPD-10754` feature flags, which controls the REST API headless selection filter management in the DSM

This has some implications:

- Data model compatibility: as this changes the way we generate data model, old models created before this FF was created (or when it was off) are no longer compatible
- Testing: some poshi tests related to picklist filters are going to break. This is a temporary situation, which we are managing under https://liferay.atlassian.net/browse/LPD-29346 in charge of migrating them


dsanz:
ci:test:sf

PR #6011, opened 2024-08-05
LPD-32891 Remove digest creation from user creation

evibudai:
Hi Team,
https://liferay.atlassian.net/browse/LPD-32891

Issue:
During user creation, we generate a digest token for WebDav, but no digest should be set when the user is created based on the expected behavior.

Solution:
Remove the logic that sets the digest.

Test fix:
https://liferay.atlassian.net/browse/LPD-1763
After the logic changed, there's no point in fixing and bringing back:
testSetDigestBeforePrerequisites
testSetEmailAndDigestBeforeScreenName
testSetScreenNameAndDigestBeforeEmailAddress

Instead, I created a test case that checks if the digest is not set after creating a user.

Please review my PR!

Thanks!


evibudai:
wrong repo

PR #6012, opened 2024-08-05
LPD-27188 Invalidate WebDAV session when the user is deactivated - follow up

stian-sigvartsen:
Sharing stian-sigvartsen/liferay-portal/pull/300 for visibility and collaboration with the team.


stian-sigvartsen:
Wrong repo

PR #6021, opened 2024-08-13
LPD-33798 Add null check for 'emails' and 'email' getters

tomibiro:


tomibiro:
opened with mistake, sorry, closing

PR #6044, opened 2024-09-06
LPD-35730 Fix A non-localizabled field is disabled in journalArticle test

gergelyszalay:
Hi,
May I ask to review this PR?

**This is for https://liferay.atlassian.net/browse/LPD-35730**, story: LPD-19627

**test name**:  A non-localized field is disabled when another translation language is selected 

**Solution**: looks like a new Metadata and fields section was introduced since, added to open Fields section if the Field is not visible

Thank you in advance!


gergelyszalay:
srry wrong repo

PR #6115, opened 2024-11-09
LPD-41441 - Removing section

marcoscv-work:
Removing no longer needed section


marcoscv-work:
gh bug, no recognise remote

PR #6137, opened 2024-12-13
LPD-42251 Create upgrade process for Web Content Display configuration

Mikellorza:
# What is this trying to solve?

https://liferay.atlassian.net/browse/LPD-42251

# How am I fixing it?

Create the upgrade process to create external reference codes for the selected journal, selected template and the group. This PR is only to review it in order to contribute to the master repository when we delete the FF since we have to made the changes of this upgrade in that moment.

# How can you verify that it works?

Running added integration tests.


Mikellorza:
ci:close

PR #6163, opened 2025-01-13
LPD-36309 Portal password policies cannot be used if LDAP password policy is enabled

rafaprax:
Ticket: https://liferay.atlassian.net/browse/LPD-36309


rafaprax:
ci:test:sf

PR #6173, opened 2025-01-21
LPD-46621 Use redirect as default value in case URL gets shortened as this is a common pattern

balazssk:
What is this trying to solve?
[LPD-46621](https://liferay.atlassian.net/browse/LPD-46621)
Paginated Document Library listing produces such a long URL, that PortletURLImpl choses to cut off backURL param from it, making View History Page's back button to be uninteractable.

How am I fixing it?
We can fall back to the redirect param. As this is a common pattern, tests are not included.

How can you verify that it works?
Follow the steps described in [LPD-46621](https://liferay.atlassian.net/browse/LPD-46621)

Cheers,
Balázs


balazssk:
ci:close

PR #6178, opened 2025-01-30
LPD-47645 Do not grant control menu acess automatically when role created

algobob:
Ticket: https://liferay.atlassian.net/browse/LPD-47645

## Motivation
We will enhance the user experience of the “Show Control Menu by Role” feature, as its current behaviour is both unintuitive and causing performance issues.

To address this, new roles will no longer be automatically added to the list of roles that can see the Control Menu. Instead, administrators will need to assign access manually as needed.

## Proposed Solution
Remove RoleModelListener class

## Steps to Verify

1. Navigate to Site Settings > Site Configuration > Menu Access. 
2. Enable Show Control Menu by Role and save the configuration.
3. Navigate to Control Panel > Roles.
4. Add a new Regular Role.
5. Return to Site Settings > Site Configuration > Menu Access.

**Current behavior:** The new role has been added to the list of roles that can see the Control Menu.
**Expected behavior**: The new role has not been automatically added to the list of roles that can see the Control Menu. Administrator will have to manually add.


algobob:
ci:test:sf

PR #6199, opened 2025-03-07
LPD-50051 - The CX is cleared from the configuration after deploy

ethib137:
https://liferay.atlassian.net/browse/LPD-50051

## Problem 

The themeCSS CX is being removed from the client extension entry relation table when it's redeployed. 

## Solution

We only want to remove themeCSS CX from the client extension entry relation table when they are "control panel" scoped. So we need to always check that before removing it. I created a private method to help with this check and also extracted a number of other instances of duplicated code to simplify it.


ethib137:
ci:test:sf

ethib137:
ci:test:relevant

ethib137:
ci:stop

PR #6247, opened 2025-04-16
LPD-53481 Removing debug step from pw test

gergelyszalay:
This is an additional commit for https://github.com/liferay-content-management/liferay-portal/pull/6564
Removing the debug step from playwright tests.

Thanks!


gergelyszalay:
wrong group

PR #6262, opened 2025-05-07
LPD-53243 Update custom-suite to add the possibility to decide if want integration

pietrobocale:
https://liferay.atlassian.net/browse/LPD-53243

- An optional -Dintegration parameter has been introduced, which can be set to true or false.

- If not specified, the default value is false.

- When integration=true, the modules-integration-postgresql155 or modules-integration-postgresql163 module is included in the batch names inside the test.properties file.

- When integration=false (or omitted), the module is not included.

@joshitagaki @nikki-pru


pietrobocale:
ci:test:sf

PR #6276, opened 2025-05-16
LPD 55708

peterrichards-lr:
Added a variable containing the fragment name to the Freemarker context and JavaScript. Both variables are called fragmentName


peterrichards-lr:
ci:test:sf

PR #6282, opened 2025-05-20
LPD-55865 Approved filter causes folder pagination issues in Web Content view

vamshikrishnamaddi:
### **What is this trying to solve?**
https://liferay.atlassian.net/browse/LPD-55865
When filtering by Approved status, and both folders and web content articles are present, the folders are getting split across multiple pages. Instead of showing all folders first, they are mixed with articles in pagination, which affects the expected order.

### **How am I fixing it?**
To fix the issue, the implementation first fetches all the folders. Then, if there's still space left on the page, it fetches only as many web content articles as needed to fill that space. Finally, it combines the folders and articles into one list, always showing the folders first.

### **How can you verify that it works?**

1. Start a Liferay instance
2. Create Six folders and  Six Web Content articles.
3. Repeat the above process
4. Apply a filter using the Approved status.

**Expected Result:**
All 12 folders should appear at the top of the content list on the first page


vamshikrishnamaddi:
“Opened in the wrong repository. Closing.

PR #6351, opened 2025-07-22
[LPD-61387] Style Book font affecting applications menu

algobob:
* Handles BOM character when providing minified css


algobob:
ci:test:sf

PR #6365, opened 2025-07-31
LPD-4641 Add alt text to images for cards in Clay Sample Web.

SelenaAungst:
https://liferay.atlassian.net/browse/LPD-4641

I resolved the errors related to alt text, but I noticed there were an additional 7 critical errors for "Form elements must have labels". I was unsure where to fix this after attempting in several places, it seems like it might be an issue with the checkbox of the card. I did notice though, it looks like it is wrapped in a label already which is one of the suggested fixes, so I was not sure about this one. 

image

@pat270

### DIFF
--- a/modules/apps/frontend-taglib/frontend-taglib-clay-sample-web/src/main/resources/META-INF/resources/partials/cards.jsp
+++ b/modules/apps/frontend-taglib/frontend-taglib-clay-sample-web/src/main/resources/META-INF/resources/partials/cards.jsp
@@ -232,6 +232,7 @@ ClaySampleImageCard claySampleImageCard = new ClaySampleImageCard();
 	>
 		<clay:image-card
 			actionDropdownItems="<%= claySampleImageCard.getActionDropdownItems() %>"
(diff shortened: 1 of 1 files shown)


pat270:
> I noticed there were an additional 7 critical errors for "Form elements must have labels".

@SelenaAungst The errors might be referring to not having any announcement associated with the checkbox. It needs to be linked with the card title with `id` and `aria-labelledby`. This probably needs to be a Clay update.

SelenaAungst:
Ah I see @pat270. In that case, maybe we can just keep this PR to fix the alt text errors and revisit the aria issues later?

SelenaAungst:
I used "thumbnail" as some of the other sample examples used this, but I do also see some examples that use the suggested format. I will change these in this case

SelenaAungst:
@pat270 I applied the recommendations, let me know if this looks good now!

ethib137:
Hey @SelenaAungst I think this should be sent to our platform experience branch, right?

SelenaAungst:
Ah correct! I resent it there, thanks!

PR #6369, opened 2025-08-07
LPD-61290 Add view action for file and content versions

Mikellorza:
https://liferay.atlassian.net/browse/LPD-61290

# What is this trying to solve?

Add view content for file and content versions. This PR add the code to add action to view version content and view version file. 

# How am I fixing it?

Modifying FDS propo transformer to add the modals.

# How can you verify that it works?

1. Go to content and file section and an entry
2. Add some versions of the entry
3. Go to the version list and click on view

# Important information

1. We don't have thumbnailURL for object entry versions and for that reason we can not se the file renderer.  I will send it in future PR's.
2. We don't have the edit_content_item struct action to render a version, so, for the moment we are showing the last version but we have to modify that struct action in order to support it. I will send it in future PR's.


Mikellorza:
ci:close

PR #6460, opened 2025-09-18
LPD-65847 - compensate for disabled field not sending correct information

wanderlast:
Hi @smottal ,

https://liferay.atlassian.net/browse/LPD-65847, tied to LPP.

The issue relates to the field.editable.domains portal properties which let us restrict who can edit certain user profile fields. When we limit it so that only administrators can edit the prefix and suffix fields the fields appear as disabled for the user when editing their own profile. This can cause an issue when a user with a prefix or suffix set saves their profile since we currently attempt to grab the value from the actionRequest. Since the fields are disabled no information related to the field is sent in the request which leads us to always assume the value for the prefix and suffix should be "0". This then causes us to fail the check for whether changes were made to the prefix/suffix since the information no longer matches the actual value that the user has set. 

I've edited the logic so that if we don't have permission to change the prefix/suffix that we always submit whatever the user currently has set as their prefix/suffix in the actionRequest. Since they don't have permission to change these two fields anyway, I believe it's safe to assume that the value should stay the same. I've also edited a near identical implementation in EditAccountUserMVCCommand so that they act the same, but if this appears unnecessary then I'll change it back. Thanks!


wanderlast:
ci:test:sf

PR #6714, opened 2025-10-27
LPD-69617

oliv-yu:


oliv-yu:
ci:test:search

oliv-yu:
ci:test:search

oliv-yu:
ci:test:search-as

pedro-oliveira446:
I checked both cases:

- The resource endpoint issue predates this change, so I think it should be handled in a follow-up ticket.
- For an AI Hub account user, the editor is read-only and Update/Save are disabled. The current editor also uses the Headless API and handles errors inline. A full-page error would only be possible through a legacy or manually crafted MVC action.

PR #7413, opened 2026-04-22
LRCI-7299 Parallelize modules-compile

kenjiheigel:
https://liferay.atlassian.net/browse/LRCI-7299

**What is being fixed:** The `modules-compile` batch in `build-test-batch.xml` runs each module's assemble task in its own `gradle-execute` call through a for-loop wrapper. That makes the batch effectively single-threaded across the axis, and modules-compile currently takes many hours across 13 nodes.

**How it is being fixed:** The `modules-compile` target now collects the full task list and dispatches it in a single `gradle-execute` call with `--parallel --continue`. The per-module JUnit reports that the old Ant-side `execute-gradle-task` macrodef used to emit are now produced by a new CI-only Gradle listener, living in `modules/init-ci.gradle` and applied from `modules/init.gradle` when `JENKINS_HOME` is set. The listener writes `modules/test-results/TEST-<dotted-module-path>.xml` in a shape byte-compatible with the Ant macro it replaces: one `<testsuite>` / `<testcase>` per module with the task duration and a `<failure>` containing the raw stacktrace on failure. Projects whose target task is skipped (typically because a dependency like `compileJava` failed) are covered by a `buildFinished` write-on-miss fallback, preserving the per-module visibility the old flow had.

### DIFF
--- a/build-test-batch.xml
+++ b/build-test-batch.xml
@@ -8712,50 +8712,17 @@ information. Make sure to commit in all format-javadoc results.
 					</then>
 				</if>
 
-				<loadresource property="modules.compile.task.groups">
+				<loadresource property="modules.compile.tasks">
 					<filterchain>
 						<replacestring from="-test:assemble" to="-test:compileTestIntegrationJava" />
-
-						<tokenfilter>
-							<replaceregex flags="g" pattern="(\S+ ){${test.module.groups.size}}" replace="\0," />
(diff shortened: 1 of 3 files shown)


kenjiheigel:
ci:test:stable

kenjiheigel:
ci:test:stable

PR #7502, opened 2026-05-18
LPD-89408 Use STANDARD cookie spec on OIDC HTTP calls

cdbm:
https://liferay.atlassian.net/browse/LPD-89408

## What is being fixed

During portal startup, Apache HttpClient emits "Invalid cookie header" WARN entries when an Identity Provider returns a `Set-Cookie` header containing an RFC 1123 `Expires` attribute (the standard "Expires=Fri, 15 May 2026 18:46:54 GMT" form). The default Apache HttpClient cookie policy is strict about cookie date formats and rejects RFC 1123 as malformed, even though the rest of the world treats it as valid. The OIDC code paths in `OAuthClientEntryLocalServiceImpl` (discovery endpoint fetch) and `OpenIdConnectHttpUtil` (token and userinfo calls) inherit that strict default and surface the warning.

## How it is being fixed

Both OIDC HTTP code paths opt into `Http.CookieSpec.STANDARD`, which parses RFC 1123 `Expires` attributes silently. The fix is a one-line `setCookieSpec` call at the two call sites. Coverage is added at two levels: a one-assertion extension to the existing `OpenIdConnectHttpUtilTest` unit test pins the enum value, and two integration tests drive the real Apache HttpClient through an embedded HTTP server returning a tracer cookie with an RFC 1123 `Expires`, asserting via `LogCapture` that the `ResponseProcessCookies` logger emits no `Invalid cookie header` WARN tagged with the tracer cookie.

### DIFF
--- a/modules/apps/oauth-client/oauth-client-persistence-service/src/main/java/com/liferay/oauth/client/persistence/service/impl/OAuthClientEntryLocalServiceImpl.java
+++ b/modules/apps/oauth-client/oauth-client-persistence-service/src/main/java/com/liferay/oauth/client/persistence/service/impl/OAuthClientEntryLocalServiceImpl.java
@@ -347,6 +347,7 @@ public class OAuthClientEntryLocalServiceImpl
 
 			Http.Options httpOptions = new Http.Options();
 
+			httpOptions.setCookieSpec(Http.CookieSpec.STANDARD);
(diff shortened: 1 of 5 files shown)


cdbm:
Wrong target repo. Reopening on liferay-appsec/liferay-portal.

PR #7536, opened 2026-05-28
LPD-92475 Stop persisting "null" for nullable AS metadata URIs

cdbm:
https://liferay.atlassian.net/browse/LPD-92475

## What is being fixed

`OAuthClientASLocalMetadataLocalServiceImpl` reduced Nimbus URIs and identifiers back to `String` with `String.valueOf`, which returns the literal string `"null"` when the source URI is null. Nimbus then round-tripped that literal through `new URI("null")` and `toJSONObject()`, so the stored metadata JSON contained entries like `"userinfo_endpoint":"null"`. On the next login, `OIDCProviderMetadata.parse` rehydrated those as non-null relative URIs and the OIDC flow blew up with "URI is not absolute" when it tried to call `toURL` on them. Rows already saved before the fix carried the same corruption and kept failing OIDC login until an admin manually re-saved each provider.

## How it is being fixed

The first two commits stop new writes from creating the corruption. Every Nimbus URI and identifier is routed through `Objects.toString(..., StringPool.BLANK)` before persisting, and each `set*EndpointURI` call in `_generateAuthorizationServerMetadataJSON` and `_generateMetadataJSON` is gated on `Validator.isNotNull` so a stray literal `"null"` still cannot reach the stored JSON even if upstream sanitization is bypassed.

The next two commits handle rows already corrupted before the runtime fix landed. A new v1_5_3 upgrade streams every `OAuthClientASLocalMetadata` row, parses both `metadataJSON` and `oAuthASMetadataJSON` via Nimbus's `JSONObjectUtils.parse`, removes any top-level entry whose value is the literal string `"null"`, and writes the pair back in a single batched UPDATE only when modified — so clean rows are byte-for-byte preserved. `Liferay-Require-SchemaVersion` is bumped to `1.5.3` and the upgrade is registered alongside the existing v1_5_2 issuer-recovery process.

### DIFF
--- a/modules/apps/oauth-client/oauth-client-persistence-test/src/testIntegration/java/com/liferay/oauth/client/persistence/internal/upgrade/v1_5_3/test/OAuthClientASLocalMetadataNullURIUpgradeProcessTest.java
+++ b/modules/apps/oauth-client/oauth-client-persistence-test/src/testIntegration/java/com/liferay/oauth/client/persistence/internal/upgrade/v1_5_3/test/OAuthClientASLocalMetadataNullURIUpgradeProcessTest.java
@@ -1,0 +1,217 @@
+/**
+ * SPDX-FileCopyrightText: (c) 2026 Liferay, Inc. https://liferay.com
+ * SPDX-License-Identifier: LGPL-2.1-or-later OR LicenseRef-Liferay-DXP-EULA-2.0.0-2023-06
+ */
+
+package com.liferay.oauth.client.persistence.internal.upgrade.v1_5_3.test;
+
+import com.liferay.arquillian.extension.junit.bridge.junit.Arquillian;
+import com.liferay.counter.kernel.service.CounterLocalService;
+import com.liferay.oauth.client.persistence.model.OAuthClientASLocalMetadata;
+import com.liferay.oauth.client.persistence.service.OAuthClientASLocalMetadataLocalService;
+import com.liferay.portal.kernel.cache.MultiVMPool;
+import com.liferay.portal.kernel.json.JSONObject;
+import com.liferay.portal.kernel.json.JSONUtil;
+import com.liferay.portal.kernel.upgrade.UpgradeProcess;
+import com.liferay.portal.test.log.LogCapture;
+import com.liferay.portal.test.log.LoggerTestUtil;
+import com.liferay.portal.test.rule.Inject;
+import com.liferay.portal.test.rule.LiferayIntegrationTestRule;
+import com.liferay.portal.upgrade.registry.UpgradeStepRegistrator;
+import com.liferay.portal.upgrade.test.util.UpgradeTestUtil;
+
+import com.nimbusds.oauth2.sdk.as.AuthorizationServerMetadata;
+import com.nimbusds.openid.connect.sdk.op.OIDCProviderMetadata;
+
+import org.junit.Assert;
+import org.junit.ClassRule;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+/**
+ * @author Christian Moura
+ */
+@RunWith(Arquillian.class)
+public class OAuthClientASLocalMetadataNullURIUpgradeProcessTest {
+
+	@ClassRule
+	@Rule
+	public static final LiferayIntegrationTestRule liferayIntegrationTestRule =
+		new LiferayIntegrationTestRule();
+
+	@Test
+	public void testUpgrade() throws Exception {
+		JSONObject brokenMetadataJSONObject = _baseMetadataJSONObject(
+		).put(
+			"authorization_endpoint", "null"
+		).put(
+			"registration_endpoint", "null"
+		).put(
+			"token_endpoint", "null"
+		).put(
+			"userinfo_endpoint", "null"
+		);
+		JSONObject brokenOAuthASMetadataJSONObject =
+			_baseOAuthASMetadataJSONObject(
+			).put(
+				"authorization_endpoint", "null"
+			).put(
+				"jwks_uri", "null"
+			).put(
+				"registration_endpoint", "null"
+			).put(
+				"token_endpoint", "null"
+			);
+
+		OAuthClientASLocalMetadata brokenOAuthClientASLocalMetadata =
+			_persistOAuthClientASLocalMetadata(
+				brokenMetadataJSONObject.toString(),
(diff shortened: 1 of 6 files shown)


modules/apps/oauth-client/oauth-client-persistence-test/src/testIntegration/java/com/liferay/oauth/client/persistence/service/test/OAuthClientASLocalMetadataLocalServiceTest.java
```diff
@@ -0,0 +1,187 @@
...
+						"token_endpoint",
+						updatedIssuer + "/protocol/openid-connect/token"
+					).toString(),
+					"openid-configuration");
+
+		String storedMetadataJSON =
+			_oAuthClientASLocalMetadata.getMetadataJSON();
+
+		Assert.assertFalse(
+			"Stored metadataJSON must not contain the literal \"null\" value " +
```
cdbm: i think there is a manual rule in /format-source against these comments

modules/apps/oauth-client/oauth-client-persistence-test/src/testIntegration/java/com/liferay/oauth/client/persistence/service/test/OAuthClientASLocalMetadataLocalServiceTest.java
```diff
@@ -0,0 +1,187 @@
...
+				"for \"userinfo_endpoint\": " + storedMetadataJSON,
+			storedMetadataJSON.contains("\"userinfo_endpoint\":\"null\""));
+
+		OIDCProviderMetadata oidcProviderMetadata = OIDCProviderMetadata.parse(
+			storedMetadataJSON);
+
+		Assert.assertNull(oidcProviderMetadata.getUserInfoEndpointURI());
+	}
+
+	private JSONObject _buildFullMetadataJSONObject() {
```
cdbm, line 102: dont think this private method is needed

modules/apps/oauth-client/oauth-client-persistence-test/src/testIntegration/java/com/liferay/oauth/client/persistence/internal/upgrade/v1_5_3/test/OAuthClientASLocalMetadataNullURIUpgradeProcessTest.java
```diff
@@ -0,0 +1,202 @@
...
+					_counterLocalService.increment());
+
+		oAuthClientASLocalMetadata.setMetadataJSON(metadataJSON);
+		oAuthClientASLocalMetadata.setOAuthASMetadataJSON(oAuthASMetadataJSON);
+
+		return _oAuthClientASLocalMetadataLocalService.
+			updateOAuthClientASLocalMetadata(oAuthClientASLocalMetadata);
+	}
+
+	private void _runUpgrade() throws Exception {
```
cdbm: check if we need this private

modules/apps/oauth-client/oauth-client-persistence-test/src/testIntegration/java/com/liferay/oauth/client/persistence/internal/upgrade/v1_5_3/test/OAuthClientASLocalMetadataNullURIUpgradeProcessTest.java
```diff
@@ -0,0 +1,202 @@
...
+			UpgradeProcess upgradeProcess = UpgradeTestUtil.getUpgradeStep(
+				_upgradeStepRegistrator, _CLASS_NAME);
+
+			upgradeProcess.upgrade();
+
+			_multiVMPool.clear();
+		}
+	}
+
+	private static final String _CLASS_NAME =
```
cdbm, line 195: check if this name is right

modules/apps/oauth-client/oauth-client-persistence-test/src/testIntegration/java/com/liferay/oauth/client/persistence/internal/upgrade/v1_5_3/test/OAuthClientASLocalMetadataNullURIUpgradeProcessTest.java
```diff
@@ -0,0 +1,202 @@
...
+public class OAuthClientASLocalMetadataNullURIUpgradeProcessTest {
+
+	@ClassRule
+	@Rule
+	public static final LiferayIntegrationTestRule liferayIntegrationTestRule =
+		new LiferayIntegrationTestRule();
+
+	@Test
+	public void testUpgrade() throws Exception {
+		JSONObject brokenMetadataJSONObject = _baseMetadataJSONObject(
```
cdbm, line 45: in the impl and in the upgrade we change more than just registration and userinfo endpoints, so why are we just testing these?

modules/apps/oauth-client/oauth-client-persistence-test/src/testIntegration/java/com/liferay/oauth/client/persistence/service/test/OAuthClientASLocalMetadataLocalServiceTest.java
cdbm: It is called twice — once here, and once from `_testAddOAuthClientASLocalMetadata` (now line 124). Leaving it.

modules/apps/oauth-client/oauth-client-persistence-test/src/testIntegration/java/com/liferay/oauth/client/persistence/internal/upgrade/v1_5_3/test/OAuthClientASLocalMetadataNullURIUpgradeProcessTest.java
cdbm: Checked sibling upgrade-process tests (`AccountRoleResourceUpgradeProcessTest`, `AccountEntryResourcePermissionUpgradeProcessTest`, `RoleResourceUpgradeProcessTest`, …) — they all use `_CLASS_NAME` for the FQCN handed to `UpgradeTestUtil.getUpgradeStep`. Leaving it to match.

PR #7548, opened 2026-06-02
LRCI-6956 Resolve the top level build URL from the build description

michaelhashimoto:
https://liferay.atlassian.net/browse/LRCI-6956

## What Is Being Fixed

The top level build URL for a Testray build was resolved only from the Testray attachment URL pattern. That left builds whose top level coordinates are recorded in the build description without a resolvable top level build URL.

## How It Is Being Fixed

TestrayBuild#getTopLevelBuildURL now first parses the build description with a MultiPattern that recognizes both the Jenkins Report anchor and the testray-results URL, extracting the master hostname, job name, and build number, and falls back to the existing Testray attachment URL matcher. TestrayImporter now writes the Jenkins Report link into the description using the Testray attachment URL for jenkins-report.html.gz (with ?authuser=0) when one is available, falling back to the Jenkins report URL. Together these let the top level build URL be recovered directly from the description.

## Why Are There No Tests?

jenkins-results-parser is CI tooling without a unit test harness for this code path; the change is verified by CI runs.


michaelhashimoto:
@brianchandotcom - This is in response to this comment:
https://github.com/brianchandotcom/liferay-portal/pull/175726#issuecomment-4597184893
I didn't agree with the public methods comment, because I'm going to be using those in following pulls.

PR #7560, opened 2026-06-12
LPD-59425 LICENSE MECHANISM WIP

jaredgorski:
WIP for license mechanism.

### DIFF
--- a/cloud/helm/default/scripts/update-license.sh
+++ b/cloud/helm/default/scripts/update-license.sh
@@ -20,7 +20,7 @@ if [ $? -ne 0 ]; then
 fi
 
 FETCHED_LICENSE_BASE64=$( \
-	echo "${LICENSE_RESPONSE}" | jq --raw-output '.license' \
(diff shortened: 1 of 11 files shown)


jaredgorski:
Gaps to carry into LCD-48603:

- **Live refresh doesn't reach running pods.** Liferay's Docker mount-to-deploy copy runs only at container startup, so a license renewed in the Secret is installed on the next pod restart, not live. Live refresh would require triggering a rollout restart on change, which conflicts with Liferay's guidance against runtime auto-deploy in multi-node setups.
- **First-boot seeding isn't deterministic.** The licenser seeds the Secret after the workload syncs, so with `LIFERAY_DISABLE_TRIAL_LICENSE=true` and an empty Secret at first boot, the initial install may need one restart cycle before the license installs. It self-heals afterward, since the license persists on the PV. A deterministic seed (initContainer or readiness gate) is the follow-up.

Both want a product decision on the refresh and seeding approach rather than a chart patch.

jpince:
ci:close

PR #7578, opened 2026-07-14
LPD-98221 Assert AI Creator checkbox state before saving

veroglez:
https://liferay.atlassian.net/browse/LPD-98221

## What Is Being Fixed

The `AICreatorInstanceSettingsPage` Playwright page object located the "Enable ChatGPT to Create Content" and "Enable DALL-E to Create Images" checkboxes with `getByText(...)` and toggled them without verifying that the toggle took effect. `getByText` matches the label's text node rather than the input control, which is a fragile way to reach a checkbox, and because no assertion followed the `check()`/`uncheck()` calls, a toggle that silently failed to register would still let the test click Save and pass.

## How It Is Being Fixed

The two checkbox locators now use `getByLabel(...)`, which resolves the actual input associated with the visible label. Each of the four toggle methods asserts the resulting checkbox state with `expect(...).toBeChecked()` / `.not.toBeChecked()` immediately after the toggle and before clicking Save, so the page object fails fast at the toggle step instead of masking the problem downstream. `expect` was added to the `@playwright/test` import to support the assertions.

## Why Are There No Tests?

This change only hardens Playwright test infrastructure — a page object used by the AI Creator settings tests. It strengthens assertions on already-covered flows and introduces no product code, so there is no new behavior to cover with additional tests.

## PR Check

**pr-check: PASS** — tested on `8075c4206c0a0e7fe2077dad09407606b55f0ff3`

| Validation | Result |
| --- | --- |
| Source Format | PASS |

### DIFF
--- a/modules/test/playwright/pages/product-navigation-applications-menu/AICreatorSettingsPage.ts
+++ b/modules/test/playwright/pages/product-navigation-applications-menu/AICreatorSettingsPage.ts
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: LGPL-2.1-or-later OR LicenseRef-Liferay-DXP-EULA-2.0.0-2023-06
  */
 
-import {Locator, Page} from '@playwright/test';
+import {Locator, Page, expect} from '@playwright/test';
 
 import {waitForAlert} from '../../utils/waitForAlert';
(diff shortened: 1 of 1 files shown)


veroglez:
ci:forward

veroglez:
ci:forward

PR #7607, opened 2026-08-19
LCD-53084: restrict crossplane-system ingress to observability + control plane

rafaelpdemelo:
The crossplane chart (2.1.3) has zero native NetworkPolicy support (no values, no templates render one under any combination), so all four policies here are hand-written via the chart's extraObjects, same mechanism already used for argocd-system in LCD-52972.

Checked the live behavior on a real cluster rather than guessing:
- Every GCP provider pod (compute, sql, storage, kms, cloudplatform, the upbound provider family) plus provider-kubernetes registers its own admission webhook on containerPort 9443 (named "webhook"), separate from crossplane core's own webhook (named "webhooks", same port). Providers carry no common label, so crossplane-webhook-ingress uses a blanket podSelector covering the whole namespace and matches the port by number (9443) since the two naming conventions disagree. This is also why a global.networkPolicy-style toggle wouldn't have worked here even if the chart had one: new providers get added over time and a name-enumerated allow list would silently miss them.
- Crossplane core calls each function runtime's grpc port (9443, confirmed via kubectl) to render compositions. crossplane-function-grpc-ingress selects function pods by the pkg.crossplane.io/function label's mere existence rather than a fixed list of function names, so it keeps covering new functions automatically.
- crossplane-metrics-ingress allows the observability namespace on the metrics port across every pod in the namespace, for the same "no shared label to narrow the selector" reason as the webhook policy.

Adds master_ipv4_cidr_block and observability_namespace to gitops/platform/variables.tf — this module didn't reference either before. master_ipv4_cidr_block defaults to 172.16.0.0/28, matching both gcp/gke/variables.tf's existing default and the real value on the infosec-rafapdemelo test cluster.

Verified with terraform validate, terraform test, and helm template against the real chart before writing the tests, then applied live via helm upgrade to the infosec-rafapdemelo sandbox cluster:
- kubectl apply --dry-run=server against an existing Function and an existing Provider both returned "configured (server dry run)", confirming the control plane can still reach both the crossplane core webhook and a provider webhook.
- A pod in a different namespace timed out reaching crossplane's metrics port by pod IP, and timed out reaching the crossplane-webhooks Service on 9443 — confirming default-deny blocks unauthorized in-cluster traffic while ipBlock still lets the real control plane through.
- All 12 pods in the namespace (core, rbac-manager, 3 functions, 7 providers) stayed Running with no restarts through the upgrade.


cloud/terraform/gcp/gitops/platform/crossplane.tf
```diff
@@ -9,6 +9,135 @@ resource "helm_release" "crossplane" {
...
+									]
+									ports=[
+										{
+											port="metrics"
+											protocol="TCP"
+										},
+									]
+								},
+							]
+							podSelector={}
```
willianmga, line 87: restrict it to pods with label `pkg.crossplane.io/provider` instead of all pods

RafaelCRC:
@rafaelpdemelo just a heads up for you to close this

PR #7624, opened 2026-09-17
 LPD-105795 Fix Geolocation map not rendering in Edge with Location Services disabled

ShrilakshmiReddy:
## What is this trying to solve?

[LPD-105795](https://liferay.atlassian.net/browse/LPD-105795) / [LPP-65464](https://liferay.atlassian.net/browse/LPP-65464)

When OS-level Location Services are disabled in Windows, creating a Web Content article with a structure containing a Geolocation field does not display the map in Microsoft Edge. Instead of showing the map centered at Latitude 0 / Longitude 0 with a marker (the expected fallback when the Google Maps provider is configured), no map is rendered at all.

## How am I fixing it?

`MapBase`'s constructor calls the browser's `navigator.geolocation.getCurrentPosition()` (via the shared `getGeolocation` helper) with no `PositionOptions`, so `timeout` defaults to the spec's ~50 day maximum — effectively infinite. Map rendering only happens inside `getCurrentPosition`'s success or error callback, via `_initializeLocation` -> `_initializeMap` -> `_createMap`; there is no other path to it.

On a normal permission denial, Chrome/Edge/Firefox invoke the error callback quickly, so the existing 0,0 fallback map renders correctly. But when the user allows the permission prompt and the underlying Windows OS-level Location Service is off, Chromium's Windows geolocation backend hangs waiting on that service instead of failing fast, so neither the success nor the error callback is ever invoked. With no timeout configured, nothing forces the fallback, and the map placeholder is left empty.

This passes a finite `timeout` (10 seconds) as the third argument to `getGeolocation()` in `MapBase.js`. That makes the browser's own timer, which runs independently of whether the OS location provider ever responds, fire the error callback after 10s if no position was returned. That callback already correctly falls back to rendering the map at 0,0, exactly as it does for a plain permission denial today.

Test coverage: a new unit test in `test/liferay/MapBase.js` asserting `getCurrentPosition` is invoked with `{timeout: 10000}`. The actual Windows/Edge hang can't be reliably reproduced in an automated test — Playwright/CI has no way to force that specific OS-level Windows behavior — so this is covered by the regression-guarding unit test plus manual verification on a Windows VM with Location Services disabled.

## How can you verify that it works?

- Disable Location Services in Windows (Settings > Privacy & security > Location > Location services)
- Go to Configuration > Site Settings > Maps > Maps and set the provider to Google Maps
- Go to Content & Data > Web Content > Structures, create a structure with a Geolocation field
- Create a new Web Content article using this structure
- When asked for the location permission, choose "Allow"
- Confirm the map now renders, centered at Latitude 0 / Longitude 0 with a marker, instead of staying blank

[LPD-105795]: https://liferay.atlassian.net/browse/LPD-105795?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LPP-65464]: https://liferay.atlassian.net/browse/LPP-65464?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ShrilakshmiReddy:
forwarded to https://github.com/liferay-frontend/liferay-portal/pull/5724

PR #7627, opened 2026-09-18
LPD-104381 Fix Category Facet vocabulary selection for asset libraries

ShrilakshmiReddy:
## What is this trying to solve?

[LPD-104381](https://liferay.atlassian.net/browse/LPD-104381)

After configuring the OOTB Category Facet with "Select Vocabularies" pointing at a vocabulary that belongs to an Asset Library (rather than a Site), the facet ignores the selection and behaves as if "All Vocabularies" were chosen — categories from every vocabulary show up in the filter instead of just the selected one.

## How am I fixing it?

The facet stores each selection as a `<owning group's ERC>&&<vocabulary's ERC>` pair. The configuration screen previously built this pairing entirely in the browser by fetching each site's vocabularies and guessing the owning group from context — which broke whenever a vocabulary actually belonged to a connected Asset Library rather than the site itself, silently dropping the selection.

This moves group discovery to the server, where it can be resolved correctly and safely:

- `CategoryFacetConfigurationDisplayContext` (new) — server-side, uses `GroupService.getUserSitesGroups()` (permission-checked, only sites the current user can see) together with `SiteConnectedGroupGroupProvider` (the same mechanism the REST taxonomy-vocabularies endpoint already uses internally) to resolve every site and its connected Asset Libraries. This is what makes it safe: visibility is gated by "can the user see this site," not by direct Asset Library membership, and it needs no REST API or schema changes at all.
- `CategoryFacetPortletConfigurationAction` — builds this display context and passes the resolved group list into the JSP.
- `configuration.jsp` — passes the `groups` list into the React component as a prop.
- `SelectVocabularies.js` — now fetches each known group's vocabularies directly (via the existing site- or asset-library-scoped taxonomy endpoints) and pairs them with that group's own already-correct external reference code, instead of inferring ownership from a broader response.

Test coverage: a new `CategoryFacetConfigurationDisplayContextTest` for the server-side group resolution, updated Jest coverage for `SelectVocabularies.js`, and a Playwright end-to-end test (`searchFacets.spec.ts`, tagged `@LPD-104381`) that replays the customer's exact repro steps.

This was developed collaboratively with @ak-ragnor, who proposed the original JS-only direction and iterated it into this server-side-resolution design after an earlier version was found to have a permission gap (it called an Asset-Library-listing endpoint gated by direct depot membership, which could hide a library's vocabularies from a Site Administrator who only had access via the site's content-sharing connection).

## How can you verify that it works?

- Create an Asset Library and connect it to the Guest site
- In the library, create two vocabularies with one category each
- Create a Basic Web Content on Guest and assign both categories
- Add Search Bar, Category Facet, and Search Results to a page
- Configure the Category Facet with "Select Vocabularies", expand the Asset Library node, and tick the first vocabulary only
- Search for the content — only the first category should be listed in the facet

[LPD-104381]: https://liferay.atlassian.net/browse/LPD-104381?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


marianoalvarosaiz:
ci:close

